PR #4, opened 2020-03-12
[#3] Add token specification

### LINKED ISSUE #3: Add contract specification
We need to agree upon the contract specification, so let's do it within this issue.

# Acceptance criteria
Contract specification is present.

awkure:
## Description

Currently specification for this smart contract is being under active discussion, so lets put this MR on hold for the time being.

## Related issue(s)

Resolves #3

- Tests
        silently reappearing again.

- Documentation
    - Haddock


docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+
+- Adds `minter` to sender's minter list to allow him mint tokens of specific
+  token type.
+
+- Sender must master minter.
+
+- Minter cannot mint tokens untill `setMintingAllowance`
+  is specified.
+
+- Cannot add minter if present in someone else's minter list.
```
kkirka: Implementation hint: it's better to have a global set (or, rather, `big_map address ()`) of all minters so that you can check this quickly without iterating over master minters' minters lists.

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+--------------------------
+
+**configureMinter**
+
+Types
+```
+tokenId = nat
+
+configureMinterParam =
+  ( address :sender
```
kkirka: Why do you need `sender` field here? Are you going to check that the transaction sender is equal to the `sender` in the parameter? In this case, you can just use `SENDER` instruction to get the sender value directly.

Or do you want an _operator_ to be able to call this function? I wonder if people would wish to offload such a sensitive action to operators.

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+- Cannot add minter if present in someone else's minter list.
+
+**removeMinter**
+
+Types
+```
+tokenId = nat
+
+removeMinter =
+  ( address :sender
```
kkirka: Same question regarding the sender as in `configureMinter`

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+- Minter address must present in sender's minter list.
+
+**setMintingAllowance**
+
+Types
+```
+tokenId = nat
+
+setMintingAllowanceParam =
+  ( address :sender
```
kkirka: Same question again (see `configureMinter`)

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+```
+(pair %setMintingAllowanceResetInterval
+  (address %sender)
+  (timestamp %interval)
+)
+```
+
+- Set the default interval of minting allowance reset.
+
+- Set to 0 to make disable minting allowance reset by stablecoin token.
```
kkirka: I propose to use `option` here but use your judgment

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+
+**mint (TBD)**
+
+Types
+```
+tokenId = nat
+
+mintParam
+  ( address :sender
+  , address :minter
```
kkirka: You can use `SENDER` to get the transaction sender and check whether it is a `minter`.
I don't get why you use `minter` here at all because AFAIU minter is the sender in all cases.

But I don't get why you do not include the token recipient in the transaction param. Who will receive the tokens in this case?

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+  given addresses.
+
+- Each minting must happen atomically, si if one of them fails, then the whole
+  operation must fail.
+
+- The operation must follow permission policies described above.
+
+- Receiving addresses must be whitelisted.
+
+- Sender must be master minter.
```
kkirka:
```suggestion
- Sender must be minter.
```

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+
+- The total amount of minted coins must not exceed the current
+  minting allowance specified for each recepient individually.
```
kkirka:
```suggestion
  minting allowance specified for each minter individually.
```

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+
+**burn (TBD)**
+
+Types
+```
+tokenId = nat
+
+burnParam
+  ( address :sender
+  , address :minter
```
kkirka: Same question about params here

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+      (pair
+        (nat %tokenId)
+        (nat %value)
+  )))
+)
+```
+
+- Decreases balances for specified wallets and their token types and the total   supply of tokens by the given amount.
+
+- Each minting must happen atomically, si if one of them fails, then the whole
```
kkirka:
```suggestion
- Each burn must happen atomically, so if one of them fails, then the whole
```

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+        (nat %tokenId)
+        (nat %value)
+  )))
+)
+```
+
+- Produces the given amounts of tokens to the wallets associated with the
+  given addresses.
+
+- Each minting must happen atomically, si if one of them fails, then the whole
```
kkirka:
```suggestion
- Each minting must happen atomically, so if one of them fails, then the whole
```

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+
+- Decreases balances for specified wallets and their token types and the total   supply of tokens by the given amount.
+
+- Each minting must happen atomically, si if one of them fails, then the whole
+  operation must fail.
+
+- The operation must follow permission policies described above.
+
+- Sender must be minter and the receiver must have a sufficient amount of
+  funds to be destroyed and be in a sender's minter list.
```
kkirka: AFAIU minter can burn his tokens _only._ Do I get something wrong? The CENTRE token spec says:
>  The minter specifies the amount of tokens to burn, and the minter must have a balance greater than or equal to the amount. Burning tokens is restricted to minter addresses to avoid accidental burning of tokens by end users.

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+
+- A minter with 0 minting allowance is allowed to burn tokens.
+
+- Contract must not be paused.
+
+**Whitelisting (TBD)**
+----------------
+
+Functions for the TZUSDC token implementation which are outside the FA1.2 Tezos Token Standard.
```
kkirka:
```suggestion
Whitelisting functions for the TZUSDC token implementation which are outside the FA2 Tezos Token Standard.
```

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+
+**Owner**
+---------
+
+**transferOnership**
+
+Types
+```
+transferOwnership =
+  ( address :owner
```
kkirka: You can use `SENDER` value and compare it against the current `owner`

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+
+**Admin**
+---------
+
+**changeTokenAdmin** address
+
+Types
+```
+changeTokenAdmin =
+  ( address :owner
```
kkirka: Same here

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+
+**Master Minter**
+---------------
+
+**changeMasterMinter**
+
+Types
+```
+changeMasterMinter =
+  ( address :owner
```
kkirka: Same here

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+
+Parameter (in Michelson):
+```
+(pair %changeMasterMinter
+  (address %owner)
+  (address %receiver)
+)
+```
+
+- Set master minter to a new address.
```
kkirka: But earlier in the spec you assumed that there would be several master minters. Have you abandoned this idea?

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+------------------------
+
+- The token(s) must be and FA2 compatible as to facilitate listing on
+  exchanges and interaction with services which support FA2
+
+**Ledger**
+==========
+
+Every address that is stored in ledger is associated with its
+current balance, current allowance and a list of minters.
```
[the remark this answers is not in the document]
awkure: The point I was making is that any "additional" storage, associated to addresses in ledger is represented by `ledgerValue`. Actually I think so too that we need to remove `allowance` value since we do not implement any `approve` method that checks for its misuse (or should we?). 

Regarding a list of minters the point was that allow for multiple addresses to create and destroy contract tokens. For more you can refer to [CENTRE specification](https://github.com/centrehq/centre-tokens/blob/78d964a1a8d481ffd8152772d7a66e47df54b3db/doc/tokendesign.md#issuing-and-destroying-tokens) document.

Regarding explicit `senders` I think that we should leave them in specification so to clarify it (as in FA2 document). The `SENDER` approach is an implementation detail and I was going to stick to it anyway.

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+
+**masterMinter**
+
+- Can add and remove minters
+
+- Can increase minting allowance for any minter
+
+- Stores the list of current minters within itself, rather
+  than exposing the map of master minter and its minters globally
+  for the whole token
```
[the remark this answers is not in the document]
awkure: Yes, I meant that each `masterMinter` has a list of it's sub-minters that can receive tokens from their parent. All minters can depend only on one `masterMinter` preventing their cyclic appearances. Yes, it needs to be clarified that it can increase allowance for any minter within its list.

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+- Can create and destroy coins (that are allowed by their current minting allowance)
+
+- Minting allowance is stored locally within the address
+  allowing for multiple minters creating and destroying tokens.
+
+**pauser**
+
+- Can pause transferring, burning and minting operations.
+  During the pause, these operations cannot be performed
+  and remain unaffected
```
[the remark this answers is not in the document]
awkure: Whops. Yes, they fail with an error if the contract is paused.

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+- Core transfer behavior may be extended. If additional constraints on tokens
+  transfer are required, FA2 token contract implementation may invoke
+  additional permission policies. If the additional permission hook fails,
+  then the whole transfer fail.
+
+- Core transfer behavior must update token balances exactly as the
+  operation parameters specify it. No changes to amount values or
+  additional transfers are allowed.
+
+- Sender must be a minter.
```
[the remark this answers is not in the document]
awkure: Actually it's not necessary)

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+
+- The amount of a token transfer must not exceed the existing token
+  owner's balance. If the transfer amount for the particular token
+  type and token owner exceeds the existing balance, then the whole
+  transfer operation fail.
+
+- Core transfer behavior may be extended. If additional constraints on tokens
+  transfer are required, FA2 token contract implementation may invoke
+  additional permission policies. If the additional permission hook fails,
+  then the whole transfer fail.
```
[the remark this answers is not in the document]
awkure: It's a part of [FA2 spec](https://gitlab.com/tzip/tzip/-/blob/76d5f3791bfbfe3c9bf95ad5ec5fc6cbeeca2d0e/proposals/tzip-12/tzip-12.md#transfer)

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+- Add or Remove token operators for the specified owners and token
+  types.
+
+- If two different commands in the list add and remove an operator
+  for the same token type, then the last one must take effect.
+
+- It's possible to update an operator for some specific tokens, or
+  to all tokens (TODO: discuss).
+
+- Sender must be minter.
```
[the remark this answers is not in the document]
awkure: At first I thought that it's be better to separate roles to minters so that we can assure that wallet has some balance before transferring. On a second thought it can be implemented easier just failing with an error if it does not have enough coins to transfer.

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+        )
+      )
+    )
+  )
+)
+```
+
+- Enables `operator` to operate on tokens held by `owner`.
+
+- `owner` and `operator` must be whitelisted.
```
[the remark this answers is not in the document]
awkure: Fair point

docs/specification.md
```diff
@@ -0,0 +1,1280 @@
...
+```
+(pair %pause
+  (address %sender)
+  unit
+)
+```
+
+- Pauses transferring, burning, minting and allowance operations so
+  that they cannot be performed. All other operations remain
+  unaffected.
```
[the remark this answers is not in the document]
awkure: And how else this can be implemented then? If you're referring to a `map (address, nat)` so that represents allowance by default. If you mean `setAllowance`-like entrypoints then I think that it's better to be paused as well as other mentioned operations, but I'm not sure.

docs/specification.md
awkure: Actually I don't think that operators should be allowed call anything but `transfer`. I pass sender address here to verify if the given address is `masterMinter` and has `minter` address inside its minter list (or rather set?)

docs/specification.md
awkure: I think it would be better to stick to FA2 spec as much as possible

docs/specification.md
awkure: Yep, it should be renamed to `:masterMinter` and `:recipient`.

docs/specification.md
awkure: Haven't touched this section at all

docs/specification.md
awkure: Nope, I'll clarify this

docs/specification.md
awkure: UPD: to check if sender address is master minter, I have already answered earlier about why I explicitly pass this in documentation even so when I'm going to use `SENDER` in actual implementation

docs/specification.md
```diff
@@ -0,0 +1,1264 @@
...
+))
+```
+
+- Removes minter from the global minter list for some specific token type
+  and sets its minting allowance to 0. Once minter is removed it will no
+  longer be able to mint or burn tokens.
+
+- Sender must be master minter.
+
+**setMintingAllowance**
```
kkirka, line 702: See https://blog.smartdec.net/erc20-approve-issue-in-simple-words-a41aaf47bca6, maybe we should consider splitting it into two functions. Not a showstopper, ofc, but nice to _not_ have this thing mentioned in the audit report in future.

docs/specification.md
```diff
@@ -0,0 +1,1264 @@
...
+Parameter (in Michelson):
+```
+(unit %acceptOwnership)
+```
+
+- Accept ownership privileges.
+
+- Sender must be a pending owner.
+
+**Admin**
```
kkirka, line 1046: Could you add "admin" to the list of roles and describe its capabilities?

docs/specification.md
awkure: IMSM we already had such solution earlier when I proposed to add `approveCAS`  method here 🤷‍♂

docs/specification.md
```diff
@@ -40,21 +40,23 @@ Below we define some roles for stablecoin contract token.
...
+- Can add and remove minters.
 
-- Can increase minting allowance for any minter
+- Can increase minting allowance for any minter if that is present in 
+  its list of minters.
 
 - Stores the list of current minters within itself, rather
   than exposing the map of master minter and its minters globally
-  for the whole token
+  for the whole token.
```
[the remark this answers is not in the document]
awkure: Yes

docs/specification.md
awkure: Actually we don't need it since due tot CENTRE specification it can only upgrade the contract which we agreed on not to for the time being

sras:
@awkure 

@gromakovsky Had asked me to review this PR. I was going through the specifications document and I had some doubts that I am including here.

>Stores the list of current minters within itself, rather than exposing the map of master minter and its minters globally for the whole token.

What does it mean by "stores within itself"? Are there a global minter list, and lists that are nested under each master minter. 

> Every address that is stored in ledger is associated with its current balance, token id and current minting allowance.

Didn't get this. Shouldn't `minting allowance` be stored for `minter` addresses only?

>The token supports the following list of user roles as is described in CENTRE Fiat Token specification 

Doesn't that doc also include an `admin` role?

> every parameter handles an additional callback call describing what needs to be done with its result.

Shouldn't this be "every parameter for query entrypoints" ?

>Removes minter from the global minter list for some specific token type and sets its minting allowance to 0.

Why need to explicitly set allowance to zero, if the minter is removed from the list? The `CENTRE Fiat Token` document also says the same.

The parameter for `mint` entrypoint seems to be

```
mintParam
  ( address :minter
  , address :recipient
  , tokenId :tokenId
  , nat     :value
  )
```

Shouldn't we use the `SENDER` instead of including the minter address in the param?

docs/specification.md
```diff
@@ -0,0 +1,1199 @@
...
+------------------------
+
+- The token(s) must be and FA2 compatible as to facilitate listing on
+  exchanges and interaction with services which support FA2
+
+**Ledger**
+==========
+
+Every address that is stored in ledger is associated with its
+current balance, token id and current minting allowance.
```
gromakovsky, line 29: > token id

The contract should hold only one type of tokens. Storing `token id` for each address does not make sense, it is always 0. Some other places should be updated as well.

docs/specification.md
```diff
@@ -0,0 +1,1199 @@
...
+
+**Token Functions**
+===================
+
+**Standard Token Functions**
+----------------------------
+
+Functions for the stablecoin token implementation which are common to the [*FA2 Tezos
+Token Standard*](https://gitlab.com/tzip/tzip/-/blob/76d5f3791bfbfe3c9bf95ad5ec5fc6cbeeca2d0e/proposals/tzip-12/tzip-12.md). The
+token implementation in general follows an [*event loop*](https://en.wikipedia.org/wiki/Event_loop) pattern meaning that every parameter
```
gromakovsky, line 84: AFAIU it only applies to getters that indeed take a callback contract. When you say that the token implementation in general follows an event loop, it makes one think that it applies to nearly everything that is going on in the contract.

Anyway, what's the purpose of this sentence? Passing a callback contract is a standard approach and basically the only way to get data from another contract in Tezos.

docs/specification.md
```diff
@@ -0,0 +1,1199 @@
...
+  )))
+)
+```
+
+- Transfers given amounts of tokens between addresses.
+
+- Each transfer must happen atomically, if one of them fails, then
+  the whole transaction fail.
+
+- Transfers must follow permission policies that are described above.
```
gromakovsky, line 123: "above"? Maybe below?

docs/specification.md
```diff
@@ -0,0 +1,1199 @@
...
+  requested token type, then the result is `false`.
+
+- It's possible to make this query for some specific tokens, or to
+  all tokens.
+
+**pause**
+
+Types
+```
+pause = (address :sender)
```
gromakovsky, line 555: Why do we pass an address as an argument instead of just using the `SENDER` instruction?

docs/specification.md
```diff
@@ -0,0 +1,1199 @@
...
+```
+
+Parameter (in Michelson)
+```
+(pair %pause
```
gromakovsky, line 560: Why is it a pair? That `unit` below is redundant.

docs/specification.md
```diff
@@ -0,0 +1,1199 @@
...
+  (address %sender)
+  unit
+)
+```
+
+- Pauses transferring, burning and minting operations so
+  that they cannot be performed. All other operations remain
+  unaffected.
+
+- Cannot be called multiple times.
```
gromakovsky, line 570: I guess you mean that it can't be called multiple times without `unpause` call between `pause` calls. But can be called multiple times during contract's lifetime.

PR #10, opened 2020-04-03
[#8] implement FA2 entrypoints

### LINKED ISSUE #8: Implement FA2 specific entyrpoints
It's decided that the contract at least follows FA2 specification, so lets implement the corresponding entrypoints within this issue.

# Acceptance criteria
Entrypoints that are described in the reference [specification](https://gitlab.com/tzip/tzip/-/blob/d3b5e954fd9117a29eb2192efdecba7ea7f3d6c5/proposals/tzip-12/tzip-12.md) are present and tested accordingly

### LINKED ISSUE #22: Use nat literals
https://github.com/tqtezos/stablecoin/blob/48e6989d1d13ec988e622bb5652a967b3c86acfb/ligo/stablecoin/spec.ligo#L16

LIGO supports `nat` literals (number followed by `n`).
Suggest to change
`const default_token_id : token_id = 0n;`

### LINKED ISSUE #23: is_operator_params -> is_operator_param
https://github.com/tqtezos/stablecoin/blob/48e6989d1d13ec988e622bb5652a967b3c86acfb/ligo/stablecoin/spec.ligo#L94

### LINKED ISSUE #24: total_supply always return a list of one element instead of mapping over each total_supply_param
https://github.com/tqtezos/stablecoin/blob/48e6989d1d13ec988e622bb5652a967b3c86acfb/ligo/stablecoin/actions.ligo#L303-L304

Although stable coin supports only one token type, it is possible that `total_supply` entry point is called with more than one `total_supply_param`. The resulting list should be the same size and an input list, even if all elements are the same.

See, correct implementation of `balance_of` entry point.

Ditto for `get_token_metadata` :
https://github.com/tqtezos/stablecoin/blob/48e6989d1d13ec988e622bb5652a967b3c86acfb/ligo/stablecoin/actions.ligo#L313

### LINKED ISSUE #25: Change entry point name from Get_token_metadata to Token_metadata
https://github.com/tqtezos/stablecoin/blob/48e6989d1d13ec988e622bb5652a967b3c86acfb/ligo/stablecoin/spec.ligo#L158

As per FA2 specification

awkure:
## Description

Within new stablecoin core some standard entrypoints should be implemented which is done in this MR.

Some known problems
- [x] `Update_operators` and `Is_operator` are not yet implemented 
- [x] `Set_minting_allowance` and `Configure_minter` are not yet unified
- [x] Documentation

## Related issue(s)

Resolves #8
Resolves #22
Resolves #23
Resolves #24
Resolves #25

- Tests
        silently reappearing again.

- Documentation
    - Haddock


gromakovsky:
Commits have `#7` in them instead of `#8`.

ligo/Makefile
```diff
@@ -6,11 +6,8 @@
...
 
 # Ligo executable
-LIGO = docker run ligolang/ligo:next
+LIGO = ligo
 
 # Compile LIGO contract into its michelson representation
 build-ligo:
-	$(LIGO) compile-contract ligo/stablecoin.ligo main
-
-test-ligo:
```
gromakovsky, line 15: There are some mentions of `test-ligo` in this file, they should be removed.

balsoft:
Hi, with `ligo:next-attempt` I get
```
$ docker run --rm -v "/tmp:/tmp" -v "$PWD":"$PWD" -w "$PWD" ligolang/ligo:next-attempt compile-contract ligo/stablecoin/core.ligo main

 If you're not sure how to fix this error, you can
 do one of the following:

* Visit our documentation: https://ligolang.org/docs/intro/what-and-why/
* Ask a question on our Discord: https://discord.gg/9rhYaEt
* Open a gitlab issue: https://gitlab.com/ligolang/ligo/issues/new
* Check the changelog by running 'ligo changelog'
ligo: : Parse error in file "core.ligo", line 1, characters 0-1, at "/":
      0: <syntax error> {}
```

Is this something LIGO guys broke or is this some update that your code doesn't confirm to?

awkure:
🤔 Actually I don't know, I've built my ligo binary from sources. You can try to add `--syntax=pascaligo` flag and see if this problem resolves. My build is relatively outdated 
```
Rolling release
Hash: 4ed4203b46ce63a76d1f608b782a80a266d67920
Date: 2020-03-09 18:24:09 +0000
CI job id:
```

balsoft:
@awkure this happens with both their old-style docker build and my nix build of latest version. I'm switching this repo to nix in #26 , and trying that on your branch yields a failure like the one above.

ligo/stablecoin/spec.ligo
```diff
@@ -0,0 +1,257 @@
...
+type operator_actions is
+| All_operator_tokens
+| Some_operator_tokens of set (token_id)
+| All_operator_tokens_except of set (token_id)
+
+type owner is address
+type operator is address
+
+type operator_storage is
+  big_map ((owner * operator), operator_actions)
```
gromakovsky, line 225: It's useless to store `operator_actions` value in case of a single token contract. An operator can operate only one type of token, that's it.

ligo/stablecoin/core.ligo
```diff
@@ -0,0 +1,24 @@
...
+  | Get_token_metadata         (params) -> get_token_metadata         (params, store)
+  | Permissions_descriptor     (params) -> permissions_descriptor     (params, store)
+  | Update_operators           (params) -> update_operators_action    (params, store)
+  | Is_operator                (params) -> is_operator_action         (params, store)
+  | Pause                      (params) -> pause                      (params, store)
+  | Unpause                    (params) -> unpause                    (params, store)
+  | Configure_minter           (params) -> configure_minter           (params, store)
+  | Set_minting_allowance      (params) -> set_minting_allowance      (params, store)
+  | Mint                       (params) -> mint                       (params, store)
+  | Burn                       (params) -> burn                       (params, store)
```
gromakovsky, line 23: Should be removed, they are not part of this issue.

gromakovsky:
I see FA2 requires all Michelson annotations to be lower-case, but in our generated contract some annotations start from capital letter (e. g. `%Some_tokens`, `%Self_transfer_permitted`). Looks like a bug.

gromakovsky:
Btw, CI complains that there are many files without licensing information.

awkure:
:coolstory: because they're with licensing information

gromakovsky:
```
reuse.project - ERROR - ligo/stablecoin/spec.ligo holds an SPDX expression that cannot be parsed, skipping the file
reuse._util - ERROR - Could not parse 'MIT *)'
```

It's actually very weird because it works for me locally (I have `reuse-0.8.1`).

gromakovsky:
You can try putting `*)` one line below :shrug:

ligo/stablecoin/spec.ligo
```diff
@@ -0,0 +1,263 @@
...
+
+(* ------------------------------------------------------------- *)
+
+(*
+ * Parameter types
+ *)
+
+type token_id is nat
+
+const default_token_id : token_id = abs (0);
```
gromakovsky, line 17:
```suggestion
const default_token_id : token_id = 0n;
```

awkure: Wtf, I clearly remember changing that, wait a minute

ligo/stablecoin/spec.ligo
```diff
@@ -0,0 +1,263 @@
...
+| Call_FA2 of parameter
+| Set_transfer_hook of set_hook_param
+
+(* ------------------------------------------------------------- *)
+
+(*
+ * Storage
+ *)
+
+type operator_actions is
```
gromakovsky, line 231: Is it used? Hopefully no.

ligo/stablecoin/spec.ligo
```diff
@@ -0,0 +1,263 @@
...
+| Some_operator_tokens of set (token_id)
+| All_operator_tokens_except of set (token_id)
+
+type owner is address
+type operator is address
+
+type operators is
+  big_map ((owner * operator), unit)
+
+// TODO: currently we need to correct our haskell tests
```
gromakovsky, line 242: I am not sure what it means, but do we still need it? Looks like a leftover. `paused` field is out of scope of this issue, but we can have it because we'll need it later. The thing I dislike is that:
1. `storage_skeleton` is called this way. Its record name is `fields`, so why not `storage_fields` then?
2. If it's supposed to have all fields beside big_maps, why doesn't it have `total_supply` and metadata?

That being said, I would just remove it.

ligo/stablecoin/actions.ligo
```diff
@@ -0,0 +1,288 @@
...
+ * tokens to be burn from.
+ *)
+function debit_from
+  ( const parameter : burn_params
+  ; const store     : storage
+  ) : storage is block
+{ var updated_ledger : ledger := store.ledger
+; case store.ledger[parameter.from_] of
+    Some (ledger_balance) -> if parameter.amount > ledger_balance
+      then failwith ("Given address has not enough balance to debit from")
```
gromakovsky, line 73: Recently error specification was added to FA2. Since it was added after we started our work, I think we can update our code later, but ideally we should follow it and fail with `INSUFFICIENT_BALANCE` here (and with `TOKEN_UNDEFINED` in `validate_token_type`).

ligo/stablecoin/operator.ligo
```diff
@@ -0,0 +1,158 @@
...
+ * Operator library for stablecoin smart-contract
+ *)
+
+#include "hooks.ligo"
+
+type owner_tokens is map (address, set (token_id))
+
+(*
+ * Validates whether the given operator parameter is equal to `SENDER`
+ * TODO: we should probably ommit `owner` in `operator_param_` as @gromak once suggested
```
gromakovsky, line 14: We can't do it because this type is part of FA2, isn't it?

ligo/stablecoin/actions.ligo
```diff
@@ -0,0 +1,288 @@
...
+
+(* ------------------------------------------------------------- *)
+
+(*
+ * Helpers and validation
+ *)
+
+(*
+ * This function fails if provided token_id is not equal to
+ * default one, restricting all operations to be monolitic
```
sras, line 18: What do you mean by `monolithic` in this context?

ligo/stablecoin/actions.ligo
```diff
@@ -0,0 +1,288 @@
...
+(*
+ * Increase the balance of an address associated with it
+ * and present in ledger. If not, create a new address with
+ * that balance given.
+ *)
+function credit_to
+  ( const parameter : mint_params
+  ; const store     : storage
+  ) : storage is block
+{ var updated_balance : nat := 0n
```
sras, line 50: Not sure but wouldn't it be less error prone to not use a `var` here? And instead use a `const` to hold the existing balance and the updated balance?

ligo/stablecoin/actions.ligo
```diff
@@ -0,0 +1,288 @@
...
+
+(*
+ * Retreive current permissions for stablecoin smart contract
+ *)
+function permission_descriptor
+  ( const parameter : permissions_descriptor_params
+  ; const store     : storage
+  ) : entrypoint is block
+{ const permissions : permissions_descriptor =
+    Layout.convert_to_right_comb ((store.permissions : permissions_descriptor_))
```
sras, line 229: It might be possible to hard code the permission descriptor here. But not yet I guess.

ligo/stablecoin/operator.ligo
```diff
@@ -0,0 +1,158 @@
...
+    , (map [] : owner_tokens)
+    )
+
+; function validate_owner
+    ( const owner  : address
+    ; const tokens : set (token_id)
+    ) : unit is block
+  { var is_operator : bool := False
+  ; if can_transfer_to_self and owner = operator
+    then skip
```
sras: Is this a bug here? if both `can_transfer_to_self` and `owner = operator` is true, then `is_operator` will remain false. Right?

awkure: Yep, refresh your page please

ligo/stablecoin/actions.ligo
awkure: That are allowed for `default_token_id`

ligo/stablecoin/spec.ligo
```diff
@@ -0,0 +1,255 @@
...
+| Is_operator            of is_operator_params
+
+(* ------------------------------------------------------------- *)
+
+(*
+ * Hooks
+ *)
+
+// TODO: currently it's the same as `transfer` but with optional
+// participants.
```
gromakovsky, line 209: That's how it is specified in FA2. Do you think we can do anything about it?

ligo/stablecoin/operator.ligo
```diff
@@ -0,0 +1,180 @@
...
+(*
+ * Remove operator from the given storage
+ *)
+function remove_operator
+  ( const param     : operator_param
+  ; const operators : operators
+  ) : operators is block
+{ validate_operator_owner_is_sender (param)
+; validate_operator_token_type (param)
+; const operator_key : (owner * operator) = (param.0, param.1.0)
```
gromakovsky, line 160: Is it necessary to use `.1.0` to access operator? Is there no better way (access by name, not by numbers)?

awkure: There's, but we should not neglet the fact that the conversion between comb tuples and actual records requires additional michelson instructions. Specifically in this case the type of `operator_key` is self-explanatory.

ligo/stablecoin/spec.ligo
awkure: No, just wanted to mention that we use it only in `validate_owner_hook`

ligo/stablecoin/actions.ligo
```diff
@@ -0,0 +1,245 @@
...
+ * if one of the parameters has different token id that is
+ * specified
+ *)
+function token_metadata
+  ( const parameter : token_metadata_params
+  ; const store     : storage
+  ) : entrypoint is block
+{ const token_metadata : token_metadata =
+    Layout.convert_to_right_comb ((store.token_metadata : token_metadata_))
+  // We drop `token_id` parameter since our contract is monolitic
```
gromakovsky, line 175: Apparently you misuse the word "monolithic". Monolithic means that all logic is implemented in one contract as opposed to the approach with `set_transfer_hook` where permission logic is handled by a separate contract.

ligo/stablecoin/actions.ligo
```diff
@@ -0,0 +1,245 @@
...
+        , (nil : list (total_supply_response))
+        )
+    , 0mutez
+    , parameter.1
+    )]
+  , store
+  )
+
+(*
+ * Returns a list of `token_metadata` calls. This is supposed
```
gromakovsky, line 163: AFAIU it always returns one call. This one call can pass a list of `token_metadata` to the callback.

ligo/stablecoin/actions.ligo
```diff
@@ -0,0 +1,245 @@
...
+  ) : entrypoint is block
+{ // We suppose that the token is already validated in `get_total_supply`
+  // and equal to `default_token_id` since our contract is monolitic
+  const total_supply_response_ : total_supply_response_ = record
+    [ token_id     = default_token_id
+    ; total_supply = store.total_supply
+    ]
+; const total_supply_response : total_supply_response =
+    Layout.convert_to_right_comb ((total_supply_response_ : total_supply_response_))
+  // We drop `token_id` parameter since our contract is monolitic
```
gromakovsky, line 141: We don't just drop it, we validate it.

ligo/stablecoin/operator.ligo
gromakovsky: Okay, that's quite sad :(

ligo/stablecoin/spec.ligo
gromakovsky: Okay, then I think it shouldn't be a TODO. Maybe just an ordinary comment.

ligo/stablecoin/operator.ligo
```diff
@@ -0,0 +1,180 @@
+// SPDX-FileCopyrightText: 2020 tqtezos
+// SPDX-License-Identifier: MIT
+
+(*
+ * Operator library for stablecoin smart-contract
+ *)
+
+#include "hooks.ligo"
+
+type owner_tokens is map (address, set (token_id))
```
gromakovsky, line 10: Why do we need such a type if we support only one `token_id`? Can't it be simplified?

awkure: I thought that's the part of specification as well 🤔

gromakovsky:
Overall all logic looks correct, but it's much more general than I expected and than written in the spec. Being more general is usually good, but in this case it may likely cause troubles due to severe gas and contract size limits. Let's see how large our contract is first.

ligo/stablecoin/spec.ligo
```diff
@@ -0,0 +1,255 @@
...
+ * Parameter types
+ *)
+
+type token_id is nat
+
+const default_token_id : token_id = 0n;
+
+(*
+ * This function fails if provided token_id is not equal to
+ * default one, restricting all operations to be monolithic
```
gromakovsky, line 21: One misuse of "monolithitic" still exists. You can say "single-token" instead.

ligo/stablecoin/actions.ligo
```diff
@@ -0,0 +1,241 @@
...
+  } with credited_storage
+
+; const upd : list (operation) =
+    generic_transfer_hook (parameters, store.permissions)
+; const ups : storage =
+    List.fold (transfer_tokens, parameters, store)
+} with (upd, ups)
+
+(*
+ * Returns a list of `balance_of` calls and leaves the storage untouched
```
gromakovsky, line 101: It's also wrong because we return only one call. Please check all the comments.

ligo/stablecoin/operator.ligo
gromakovsky: It's not part of the specification, apparently this type is only used in `validate_operators` and you put `transfer_operation.1.1.0` into its `set (token_id)` which must be `0n`. So this set always contains one item which is `0n`.

gromakovsky: Then you pass it to `is_operator_impl` which does `validate_operator_token_type`, but token type is also validated in `transfer_tokens`, so you validate twice.

ligo/stablecoin/actions.ligo
```diff
@@ -98,7 +98,10 @@ function transfer
 } with (upd, ups)
 
 (*
- * Returns a list of `balance_of` calls and leaves the storage untouched
+ * Retreives a list of `balance_of` calls that is specified for multiple
```
gromakovsky, line 101: The right spelling is "Retrieves", it should be fixed in many places. And I still don't understand why you say "a list of calls". I assume by "call" you mean `TRANSFER_TOKENS` operation and this function returns just one call. I mean, technically it's still a list (with one item), but imo this "list" is confusing.

ligo/stablecoin/operator.ligo
```diff
@@ -0,0 +1,177 @@
...
+; function validate_params
+    ( const param : transfer_param
+    ) : unit is block
+  { var is_operator : bool := False
+  ; const owner : owner = param.0
+  ; if can_transfer_from_self and owner = operator
+    then { is_operator := True }
+    else
+        { const operator_param : operator_param = (owner, (operator, (M_right (set [param.1.1.0]) : operator_tokens)))
+        ; is_operator := is_operator_impl (operator_param, operators)
```
gromakovsky, line 113: But it's still quite inefficient, isn't it? You still validate token_id in `is_operator_impl`, so `transfer` validates it twice. Can't you just do `Big_map.find_opt` here? I don't see a big benefit in extracting `is_operator_impl` here :shrug:

gromakovsky:
It does not resolve #39, there are other changes not taken into account here (and the spec should be updated).

awkure:
No, it does not fail, it's me who've uploaded tests before pushing to this MR

ligo/stablecoin/spec.ligo
```diff
@@ -0,0 +1,254 @@
...
+type token_metadata_params_ is record
+  token_ids : list (token_id)
+; callback  : contract (list (token_metadata_response))
+end
+
+type token_metadata_params is michelson_pair_right_comb(token_metadata_params_)
+
+type operator_tokens_ is
+| All_tokens
+| Some_tokens of set (token_id)
```
sras, line 110: @awkure Is this used still ?

awkure: Good catch, no, it's unused

sras:
Tests pass for me as well.

gromakovsky:
AFAIU nothing stops us from merging this PR. @awkure please do that unless you have good reasons not to do that.

PR #12, opened 2020-04-08
[#3] Update specification: part1

gromakovsky:
## Description

Many things need to be updated in the specification. This PR contains the most obvious fixes that I think we have agreed upon.

## Related issue(s)

Resolves part of #3

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -79,8 +79,8 @@ Below we define some roles for stablecoin contract token.
 **Token Functions**
 ===================
 
-**Standard Token Functions**
-----------------------------
+**Standard FA2 Token Functions**
+--------------------------------
 
 Functions for the stablecoin token implementation which are common to the [*FA2 Tezos
 Token Standard*](https://gitlab.com/tzip/tzip/-/blob/76d5f3791bfbfe3c9bf95ad5ec5fc6cbeeca2d0e/proposals/tzip-12/tzip-12.md). The
@@ -447,58 +447,6 @@ Parameter (in Michelson)
 - Since the contract supports only a single token type, `tokenId` must be 0.
   It is passed because FA2 requires that.
 
-**approveOperatorUpdate (TODO)**
-
-Types
-```
-tokenId
-
-operatorTokens =
-  | AllTokens
-  | SomeTokens (set tokenId)
-
-operatorParam =
-  ( address        :owner
-  , address        :operator
-  , operatorTokens :tokens
-  )
-
-approveOperatorUpdate = operatorParam
-```
-
-Parameter (in Michelson):
-```
-(pair %approveOperatorUpdate
-  (pair %operator
-      (address %owner)
-      (pair
-        (address %operator)
-        (or %tokens
-          (unit %allTokens)
-          (set %someTokens nat)
-        )
-    ))
-  (contract %callback
-    (pair %operator
-      (address %owner)
-      (pair
-        (address %operator)
-        (or %tokens
-          (unit %allTokens)
-          (set %someTokens nat)
-        )
-      )
-    )
-  )
-)
-```
-
-- Enables `operator` to operate on tokens held by `owner`.
-
-- `owner` must be whitelisted.
-
-- Contract must not be paused.
-
 **isOperator**
 
 Types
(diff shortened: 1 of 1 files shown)


docs/specification.md
```diff
@@ -569,6 +517,65 @@ Parameter (in Michelson):
 - Since the contract supports only a single token type, `tokenId` must be 0.
   It is passed because FA2 requires that.
 
+**Custom (non-FA2) token functions**
+====================================
+
+Functions for the stablecoin token implementation not present in FA2, but related to token transfers.
+
+**approveOperatorUpdate (TODO)**
```
awkure, line 521: Also we should fix its type to be a unit since we don't bother setting operators for multiple token types

docs/specification.md
```diff
@@ -227,7 +233,10 @@ Parameter (in Michelson)
 
 - Returns total supply for multiple tokens. It accepts a list of
   token ids and a callback which accepts a list of
-  `getTotalSupplyResponse` which is conequently passed to it.
+  `total_supply_response` which is conequently passed to it.
+
+- Since the contract supports only a single token type, `tokenId` must be 0.
+  It is passed because FA2 requires that.
 
 **getTokenMetadata**
```
awkure, line 241: I suppose this will be unused due to single-token nature of our contract?

awkure:
Oh no, trying to reword last commit message messes up the whole commit history, thanks github, very cool.

docs/specification.md
```diff
@@ -235,39 +235,39 @@ Parameter (in Michelson)
...
-getTokenMetadataParam =
-  ( list tokenId                             :tokenIds
-  , contract (list getTokenMetadataResponse) :callback
+get_token_metadataParam =
+  ( list token_id                               :token_ids
+  , contract (list get_token_metadata_response) :callback
   )
 
-getTokenMetadata = getTokenMetadataParam
+get_token_metadata = get_token_metadataParam
```
gromakovsky, line 260: What about this `Param`?

docs/specification.md
gromakovsky: `unit`? Type of `approveOperatorUpdate`? What will it do then?

docs/specification.md
gromakovsky: It can be used to get metadata about this single token. And we must have an entrypoint with exactly this type because it's required by FA2.

docs/specification.md
awkure: Once is called sender will be defined in transfer's ledger value as operator

gromakovsky: Removed this function as explained in the last commit.

docs/specification.md
```diff
@@ -681,14 +648,11 @@ Parameter (in Michelson)
...
-    (address %minter)
-    (nat %tokenId)
-))
+  (address %minter)
+)
 ```
 
-- Adds `minter` to global minter list to allow him mint tokens of specific
-  token type.
+- Adds `minter` to global minter list to allow him mint tokens.
```
sras, line 613: @gromakovsky AFAIU adding to the global minter list does not allow him to mint tokens. For that the mint allowance has to be set to some non-zero value in the ledger as well.

docs/specification.md
```diff
@@ -746,14 +702,11 @@ Parameter (in Michelson)
...
-)))
+    (nat %amount)
+))
 ```
 
-- Set the amount of allowed minting allowance for an address with some
-  specific token type.
+- Set the amount of allowed minting allowance for an address.
 
 - Minter must present in sender's minter list.
```
sras, line 669: But not necessarly in the global minter list?

docs/specification.md
```diff
@@ -940,14 +850,12 @@ Parameter (in Michelson):
...
-      (nat %tokenId)
-      (nat %value)
-  ))
+    (nat %value)
+  )
 )
 ```
 
-- Decreases balances for senders and their token types and the total supply of tokens by the given amount.
+- Decreases balances for senders and the total supply of tokens by the given amount.
```
sras, line 810: Might also mention that

> Burning tokens will not increase the mintingAllowance of the address doing the burning.

https://github.com/centrehq/centre-tokens/blob/78d964a1a8d481ffd8152772d7a66e47df54b3db/doc/tokendesign.md#burning

gromakovsky: Yes, will add.

docs/specification.md
gromakovsky: Well, below we are saying that:
> - Minter cannot mint tokens untill `set_minting_allowance`  is specified.

I think wording can be a bit improved here. Adding to the minter list is required, but not sufficient. Will improve in the next part.

docs/specification.md
gromakovsky: I don't touch it in the first part yet. Currently I don't understand the situation with minter lists. I will update it in the next part.

PR #14, opened 2020-04-09
[#3] Update the specification (part 2)

gromakovsky:
## Description

The contract specification still has a lot of inconsistencies with the actual requirements. This part has less obvious changes.

[Rendered](https://github.com/tqtezos/stablecoin/blob/gromak/specification2/docs/specification.md)

There will be one more part (hopefully only one). Plans for that part:
1. Figure out specifics regarding FA2 entrypoints. My current understanding is that FA2 is very general and our contract is, obviously, specific, so we can specify various things more precisely.
2. Maybe write entrypoint arguments more concisely (use `view` synonym, tuple syntax, specify how pseudocode is written).
3. Proofread everything, most likely some other fixes will be necessary.

## Related issue(s)

Relates to #3 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -21,6 +21,22 @@ These specifications were assembled with the following references:
 
 - The token contract must store tokens of a single type.
 
+# Differences with CENTRE
+
+The stablecoin contract is based on the CENTRE fiat token design, but diverges from it in several aspects:
+1. CENTRE token implements the ERC-20 interface.
+Tezos version of it is FA1.2.
+The stablecoin contract does not implement the FA1.2 interface, it implements the FA2 interface.
+2. Since the names in FA2 are snake\_cased, the names in the stablecoin contract also snake\_cased.
+3. FA2 uses the word "owner" to refer to the address that owns some tokens.
+At the same time, CENTRE calls the central entity who can manage roles the "owner".
+It leads to ambiguity, but the meaning of the word "owner" should be clear from the context.
+It is not a difference per se, but due to this ambiguity we may rename the "owner" role.
+4. `mint` and `burn` entrypoints take lists of pairs/amounts respectively.
+They follow FA2 style where most entrypoints operate on lists.
+In CENTRE they take a single item (a pair or an amount).
+5. TODO: whitelisting and blacklisting.
+
 # Ledger
 
 Every address that is stored in ledger is associated with its
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -10,7 +10,7 @@ These specifications were assembled with the following references:
 - [*CENTRE Fiat Token design*](https://github.com/centrehq/centre-tokens/blob/78d964a1a8d481ffd8152772d7a66e47df54b3db/doc/tokendesign.md) which is served as a reference specification to tezos stablecoin smart contract.
 
 - Tezos Token Standard:
-  [*FA2*](https://gitlab.com/tzip/tzip/-/blob/76d5f3791bfbfe3c9bf95ad5ec5fc6cbeeca2d0e/proposals/tzip-12/tzip-12.md)
+  [*FA2*][FA2]
 
 - [*Michelson Contract Interfaces and Conventions*](https://gitlab.com/tzip/tzip/blob/ae2f1e7ebb3454d811a2bea3cd0698b0e64ccea5/proposals/tzip-4/tzip-4.md) TZIP which defines `view` and `void` type synonyms
 
@@ -114,8 +114,8 @@ Each entrypoint MUST be callable using the default entrypoints machinery of Mich
 
 ## Standard FA2 Token Functions
 
-Functions for the stablecoin token implementation which are common to the [*FA2 Tezos
-Token Standard*](https://gitlab.com/tzip/tzip/-/blob/76d5f3791bfbfe3c9bf95ad5ec5fc6cbeeca2d0e/proposals/tzip-12/tzip-12.md).
+Functions for the stablecoin token implementation which are common to the
+[*FA2 Tezos Token Standard*][FA2].
 Some entrypoints (e. g. getters) follow an [*event loop*](https://en.wikipedia.org/wiki/Event_loop)
 pattern meaning that their arguments have an additional callback call
 describing what needs to be done with its result.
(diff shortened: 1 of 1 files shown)


docs/specification.md
```diff
@@ -21,6 +21,19 @@ These specifications were assembled with the following references:
...
+1. How many addresses can be assigned to each role?
+We have made some assumptions, see the [Roles](#roles) section below, please check them.
+For non-unique roles: is there an upper limit on the number of addresses?
+2. Can one address have more than one role?
+3. Do we want to rename the "owner" role to something different given that FA2 also uses this word?
+4. Are any getter entrypoints not present in FA2 required?
+Should they be `void` or `view`?
+Note that the contract will have storage annotations and each getter entrypoint increases contract's size and gas costs.
+At this point it's hard to predict whether we will have issues with the contract's size and gas costs since we've never used LIGO, but it's quite likely that we will.
+5. Should the contract reject transfers with non-zero XTZ AMOUNT (when someone calls the contract with non-zero amount)?
```
awkure, line 35: Why shouldn't?

docs/specification.md
```diff
@@ -21,6 +21,19 @@ These specifications were assembled with the following references:
 
 - The token contract must store tokens of a single type.
 
+# Questions
+
+1. How many addresses can be assigned to each role?
```
awkure, line 26: Well, I think that FA2 spec clearly says that each role has exactly one address associated to it, otherwise we face a lot of consensus (for instance multiple master minters that make a cyclic dependency to one another allowing for generation of infinite amount of coins) problems that aren't mentioned anywhere

docs/specification.md
```diff
@@ -21,6 +21,19 @@ These specifications were assembled with the following references:
...
+We have made some assumptions, see the [Roles](#roles) section below, please check them.
+For non-unique roles: is there an upper limit on the number of addresses?
+2. Can one address have more than one role?
```
awkure, line 29: I don't think so due to the problems I mentioned earlier, but who knows what guys from TQ have in mind

docs/specification.md
gromakovsky: I don't know, but it's not explicitly stated anywhere, so it's better to confirm that it's needed. And if it is not strictly necessary, we will be able to throw it away if we hit gas issues and contract size issues.

docs/specification.md
gromakovsky: FA2 does not know anything about roles like "master minter". Where does it say that each has exactly one address?

awkure: Because it leads to a whole bunch of problems that I described above and that in any case should be mentioned in FA2. By looking at its example implementation I can say that it was stated to use only one.

sras: @gromakovsky I think it is coming from https://github.com/centrehq/centre-tokens/blob/master/doc/tokendesign.md

If you check the section "Reassigning Roles", the content there seem to imply that each role, except that of 'minters' can be assigned to only one address at a time.

docs/specification.md
```diff
@@ -587,87 +596,47 @@ Parameter (in Michelson)
...
-- Adds `minter` to global minter list to allow him mint tokens.
+- Adds `minter` to the global minter list to allow him to mint tokens (if `minter` is not in the list already).
 
-- Sender must be master minter.
+- Sets the specified minting allowance (`new_minting_allowance`) for this minter.
 
-- Minter cannot mint tokens untill `set_minting_allowance`
-  is specified.
+- `current_minting_allowance` must be explicitly passed.
+If it does not match the actual minting allowance, the transaction MUST fail.
```
sras, line 625: > If it does not match the actual minting allowance

Or if the minter does not exist already, right?

docs/specification.md
```diff
@@ -587,87 +596,47 @@ Parameter (in Michelson)
...
-    (address %minter)
-    (nat %amount)
-))
-```
-
-- Set the amount of allowed minting allowance for an address.
-
-- Minter must present in sender's minter list.
-
 **get_minting_allowance**
```
sras, line 640: What is the required behavior if an address is not in the global list of minters, but actually holds some minting allowance in the ledger?

docs/specification.md
gromakovsky: Yes, but
1. "seem to imply" does not make me 100% sure, it's better to ask explicitly.
2. Maybe they can be assigned to at most one address rather than exactly one address. E. g. maybe it's ok to have no pausers.

docs/specification.md
gromakovsky: > What is the required behavior if an address is not in the global list of minters, but actually holds some minting allowance in the ledger?

How is it possible? `removeMinter` sets minting allowance to 0, that's written in our document and in CENTRE document. But we can specify that `get_minting_allowance` returns 0 for addresses that are not minters.

docs/specification.md
gromakovsky: I think it's simpler to not distinguish removed minters and minters with 0 minting allowance. I don't see how it can be useful to distinguish them. But I'll add it to the list of questions just to be sure.

gromakovsky: Oh wait, there is important difference: a minter with 0 minting allowance can burn tokens and removed minter can't :thinking:

gromakovsky: Okay, then I guess it's cleaner to change `current_minting_allowance` to `option nat` and treat `None` as "not a minter".

gromakovsky:
@sras I've updated both entrypoints that you commented. Now `option nat` represents current minting allowance and `None` means that the minter is deleted.

docs/specification.md
sras: Ah yes. I somehow managed to miss that.

docs/specification.md
sras: Ok, so does that mean that we can ditch the global minter list now, because it seems to be redundant?

gromakovsky: Well, "global minter list" is now present in only two places: `configure_minter` and `remove_minter`.

The CENTRE doc also says "the list of minters":
> This will remove the minter from the list of minters…

I guess I'll just remove the word "global" because it is indeed redundant now (there are no "local" lists of minters).

gromakovsky: Conceptually we should maintain a list of minters. We don't have to store them literally in a `list`, in practice we will probably store them in a `map` or a `big_map`. But from end user's perspective there is still a list.

PR #16, opened 2020-04-13
[#15] Investigate whether we can evaluate code lazily if needed

### LINKED ISSUE #15: Investigate whether we can load code lazily if needed

The contract we are going to implement is not quite small and we are very likely to hit some limits that Tezos has, namely gas limit and operation size limit. There are at least three ways to resolve such a problem if it appears:
1. Make some local optimizations.
2. Throw away some code (e. g. unnecessary getters).
3. Load code lazily.

The first two options are quite limited. We may be able to throw away something, but we shouldn't rely on that in any case (obviously). So chances are that we will have to use the third option.

Probably the best way to do that is to have such storage:
```
data Storage =
  { entrypoint1 :: BigMap () (Lambda (param1, Storage') ([Operation], Storage'))
  , entrypoint2 :: BigMap () (Lambda (param2, Storage') ([Operation], Storage'))
  ...
  , storage :: Storage'  -- actual data
  }
```

In this case all code remains typed, but is stored in `big_map`s and is loaded lazily, only when we use it. In practice, each contract call will load only one entrypoint and it should save a lot of gas. Also it should decrease origination operation size because we can fill these `big_map`s after origination.

Currently we need to figure out whether this approach is possible to implement in LIGO.

# Acceptance criteria

Implement a contract that stores one integer and takes a `or (pair int int) int`. In the first case it puts the sum of pair's items to the storage, in the latter case it puts the `int` itself. It should be implemented using the aforementioned pattern. In LIGO of course. Just a proof of concept.

**OR**

If you find it impossible, explain why.

awkure:
## Related issue(s)

Resolves #15 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/test.tz
+++ b/test.tz
@@ -1,0 +1,25 @@
+{ parameter
+    (big_map
+       bool
+       (pair (lambda %0
+                (pair (pair %0 (int %0) (int %1)) (int %1))
+                (pair (list %0 operation) (int %1)))
+             (pair %1 (int %0) (int %1)))) ;
+  storage int ;
+  code { DUP ;
+         CAR ;
+         DIP { DUP ; CDR } ;
+         PAIR ;
+         DUP ;
+         CAR ;
+         PUSH bool True ;
+         GET ;
+         IF_NONE { PUSH string "memems" ; FAILWITH } { DUP ; DIP { DROP } } ;
+         DUP ;
+         CDR ;
+         DIP { DIP { DUP } ; SWAP ; CDR } ;
+         PAIR ;
+         DIP { DUP ; CAR } ;
+         EXEC ;
+         DIP { DROP 3 } } }
+
--- a/test.tz
+++ b/test.tz
@@ -1,25 +1,78 @@
-{ parameter
-    (big_map
-       bool
-       (pair (lambda %0
-                (pair (pair %0 (int %0) (int %1)) (int %1))
-                (pair (list %0 operation) (int %1)))
-             (pair %1 (int %0) (int %1)))) ;
-  storage int ;
+{ parameter (or (int %put) (pair %sum (int %0) (int %1))) ;
+  storage
+    (pair (pair (big_map %put_entry
+                   bool
+                   (lambda (pair (int %0) (int %1)) (pair (list %0 operation) (int %1))))
+                (int %state))
+          (big_map %sum_entry
+             bool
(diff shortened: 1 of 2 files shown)


awkure:
With current implementation the execution of seemingly correct commands fail with following error log
```
$ ligo run-function test.ligo main "(Sum (big_map [(unit) -> (sum, (1, 2))]), 2)"                                                     [3ebc4e8]
ligo: error parsing program code
$ ligo compile-contract test.ligo main                                                                                              
ligo: not a comparable type: unit
```

Note that the contract itself parses correctly (for instance if we define dummy `unit -> unit` lambda and execute it from cli).

gromakovsky:
> ligo: not a comparable type: unit

Well, it makes sense because `unit` is not comparable indeed. Let's replace it with `bool` then :man_shrugging: 

So please do it and then share generated Michelson contract.

awkure:
But semantically it should be because we have to have a unit for comparable semigroup in any case

test.tz
```diff
@@ -0,0 +1,25 @@
+{ parameter
+    (big_map
```
gromakovsky, line 2: No, there should be 1 big_map per entrypoint. So in this case there should be 2 big_maps.

awkure:
Somewhat strange seeing the following error when trying to execute it 
```
$ ligo run-function test.ligo main "(Sum (1,2), record [sum_entry = big_map [(True) -> sum]; put_entry = big_map [(True) -> put]; state=0])" 
ligo: generated application. different kinds:  {"a":"int","b":"record[put_entry -> (TO_Big Map (bool,( int * int ) -> ( (TO_list(operation)) * int ))) , state -> int , sum_entry -> (TO_Big Map (bool,( ( int * int ) * int ) -> ( (TO_list(operation)) * int )))]"}
```

Meaning that the ligo compiller somewhat skips the second argument in passed record and somewhat substitutes expected `state` with passed `put_entry`.

gromakovsky:
Is it possible to print lambdas that should be put into storage?

gromakovsky:
As a backup plan we can even write this template in Lorentz and implement business logic in LIGO. Then we will print lambdas from LIGO code and put them into the storage of the template contract written in Lorentz.

awkure:
Uhm, no. There's no any printing capability currently available in LIGO. It prints values only when it returns some.

gromakovsky:
See https://github.com/tqtezos/stablecoin/issues/15#issuecomment-635437548

PR #21, opened 2020-04-23
Specify whitelisting/blacklisting

gromakovsky:
## Description

We've got some information about whitelisting/blacklisting in stablecoin. This PR updates the specification to reflect this information. Additionally, I have slightly updated `FA2 Specifics`.

## Related issue(s)

#3 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -91,14 +91,18 @@ In CENTRE they take a single item (a pair or an amount).
 5. `configure_minter` takes an additional argument – current minting allowance – to prevent front-running attacks.
 6. CENTRE does not have `update_operator`, the closest analog is `approve`, but it does not explicitly say whether `approve` is paused by `pause`.
 In this contract `update_operator` IS paused.
-7. TODO: whitelisting and blacklisting.
+7. CENTRE defines the `blacklister` role and blacklisting is part of the CENTRE Fiat Token contract.
+Our project is different: whitelisting and blacklisting are offloaded to a separate [`Safelist` contract](#safelist).
 
 # State model
 
 This chapter provides a high-level overview of the contract's state.
 Note that the actual storage type is implementation detail and is not specified here.
-The contract maintans a ledger of addresses and manages some special roles as described below.
-Token operations can be paused, so the contract also knows whether it is paused.
+
+* The contract maintains a ledger of addresses and manages some special roles as described below.
+* Token operations can be paused, so the contract also knows whether it is paused.
+* Token operations can be guarded by the [`Safelist` contract](#safelist), so the contract also optionally stores an address of such contract.
+If its value is `None`, no safelist checks are performed.
 
 ## Roles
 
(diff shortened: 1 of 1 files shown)


docs/specification.md
```diff
@@ -49,13 +51,12 @@ The "transfer hook" approach is recommended, but may be infeasible due to gas co
 3. We will not implement any `owner_hook`.
 The contract is usable without them because we stick to `Optional_owner_hook`.
```
sras, line 52: I didn't understand this part. Isn't owner hook interface part of the receivers (contracts) of the USDC token? 
AFAIU optional owner hook just means that a transfer won't fail if the sender/reciever contracts does not implement an owner hook.

docs/specification.md
```diff
@@ -90,14 +91,18 @@ In CENTRE they take a single item (a pair or an amount).
 5. `configure_minter` takes an additional argument – current minting allowance – to prevent front-running attacks.
 6. CENTRE does not have `update_operator`, the closest analog is `approve`, but it does not explicitly say whether `approve` is paused by `pause`.
 In this contract `update_operator` IS paused.
-7. TODO: whitelisting and blacklisting.
+7. CENTRE defines the `blacklister` role and blacklisting is part of the CENTRE Fiat Token contract.
+Our project is different: whitelisting and blacklisting are offloaded to a separate [`Safelist` contract](#safelist).
```
sras, line 95: Isn't this kind of functionality meant to be implemented by the "transfer_hook" FA2 spec and "set_transfer_hook" infra?

docs/specification.md
gromakovsky: > AFAIU optional owner hook just means that a transfer won't fail if the sender/reciever contracts does not implement an owner hook.

Yes. I. e. we let each user decide whether they want to implement a hook or not. So if Alice wants to do something when funds are transferred to her, she will do it. If not, she won't do it and she can use a regular `tz*` address. That's what I mean by saying that the contract is usable. So even if we don't implement any owner_hook, anyone can take this contract and use it. If made owner hook mandatory, the contract wouldn't be usable without them.

docs/specification.md
gromakovsky: It's not specified whether we will implement a monolithic contract or a modular one with a transfer hook. If we implement a monolithic one, it will call safelist contract directly. If we implement a modular one, it will call the transfer hook which will in turn call the safelist contract.

docs/specification.md
```diff
@@ -90,14 +91,18 @@ In CENTRE they take a single item (a pair or an amount).
...
 
 # State model
 
 This chapter provides a high-level overview of the contract's state.
 Note that the actual storage type is implementation detail and is not specified here.
-The contract maintans a ledger of addresses and manages some special roles as described below.
-Token operations can be paused, so the contract also knows whether it is paused.
+
+* The contract maintans a ledger of addresses and manages some special roles as described below.
```
awkure, line 102:
```suggestion
* The contract maintains a ledger of addresses and manages some special roles as described below.
```

docs/specification.md
sras: I think the confusing part is "We will not implement any `owner_hook`.". It would have made sense to me if it said "We will not require any owner_hook". But I get what you mean.

docs/specification.md
sras: I get it now.

PR #34, opened 2020-05-13
[#13] Add FA2 tests and Nettest tests (Pure and via tezos-client)

### LINKED ISSUE #13: Implement tests for FA2 entrypoints
Implement tests to check FA2 entrypoints and expected behavior.

## Acceptance criteria

All the FA2 entrypoints and their behaviors should be covered in the tests. (I plan to add more details on specifically what will be included)

sras:
## Description

Adds tests for FA2 interface.

## Related issue(s)

#13 

Resolves #13 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


gromakovsky:
But currently there is no way for CI to pass here because there are no changes from #10. Can you rebase it on #10 branch and change this PR's target to that branch?

haskell/test/Stablecoin.hs
```diff
@@ -0,0 +1,94 @@
...
+
+spec_StablecoinNettest :: Spec
+spec_StablecoinNettest = specWithUntypedContract "test/resources/stablecoin.tz" nettest_scenario
+
+spec_StablecoinNettestReal :: Spec
+spec_StablecoinNettestReal =
+  specWithUntypedContract "test/resources/stablecoin.tz" $ real_nettest_StablecoinNettest
+
+real_nettest_StablecoinNettest :: U.Contract ->  Spec
+real_nettest_StablecoinNettest contract = describe "Stablecoin real network behavior" $
```
gromakovsky, line 38: We normally put it into a separate test-suite, because it requires some extra configuration (`tezos-client` that talks to a real node) and CLO parsing.

sras:
@gromakovsky Have rebased on #10  and disabled running of nettest test via client.

haskell/test/Stablecoin.hs
sras: I have disable that test for now.

gromakovsky: Well, in the end it should be a separate test-suite anyway because it requires custom CLO parsing. We shouldn't hardcode `MorleyClientConfig`, that's bad UX (e. g. no way to change `verbose` flag).

Probably you can try to integrate it with `tasty` so that it somehow parses its own command line options, but that seems harder than putting into a separate test-suite.

haskell/src/Lorentz/Contracts/Stablecoin.hs
```diff
@@ -2,7 +2,91 @@
...
+  ) where
+
+import Lorentz
+import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
+import Lorentz.Contracts.Spec.FA2Interface (OperatorTransferMode(..), OwnerTransferMode(..), SelfTransferMode(..))
+import qualified Lorentz as L
+import Util.Named
+
+-- Parameter
+type Parameter = FA2.Parameter
```
gromakovsky, line 18: Btw, it's not the actual parameter our contract has. In `stablecoin.tz` the first (topmost) entrypoint is `balance_of`, but here it is `transfer`. It works because we never use `Parameter` directly, apparently we only use it as a type parameter for `TAddress` to satisfy some entrypoint-related constraints. I think it's ok, but should be mentioned in a comment (that it's not the actual parameter, rather some isomorphic one).

haskell/src/Lorentz/Contracts/Stablecoin.hs
```diff
@@ -2,7 +2,91 @@
...
+
+-- Permissions descriptor
+data OwHookOptReq = OptOH () | ReqOp ()
```
gromakovsky, line 21: Are these `()`s required? I thought they are implied automatically if they are missing :thinking:

haskell/test/Stablecoin.hs
```diff
@@ -0,0 +1,107 @@
...
+module Stablecoin
+  ( spec_Stablecoin
+  , spec_StablecoinNettest
+  , stablecoinNettestReal
+  ) where
+
+import Test.Hspec
+
+import Tezos.Core
+import Data.Maybe
```
gromakovsky, line 13: It should be in another group and be explicit (`Test.HSpec` as well).

haskell/test/Nettest.hs
```diff
@@ -0,0 +1,57 @@
+-- SPDX-FileCopyrightText: 2020 tqtezos
+-- SPDX-License-Identifier: MIT
+
+module Nettest
+  ( scTransferScenario
+  )where
```
gromakovsky, line 6:
```suggestion
  ) where
```

gromakovsky:
Nettest running will be done separately later in #37.

haskell/test/Stablecoin.hs
```diff
@@ -0,0 +1,107 @@
...
+    NetTest.runNettestClient env (scTransferScenario (fromJust . mkInitialStorage) contract)
+
+nettest_scenario :: U.Contract  -> Spec
+nettest_scenario contract = describe "Stablecount nettest (integrational)" $
+  it "can be originated and execute a transfer call" $
+    integrationalTestExpectation $
+    NetTest.nettestToIntegrational $
+    scTransferScenario (fromJust . mkInitialStorage) contract
+
+originationRequestCompatible :: OriginationParams -> Bool
```
awkure, line 52: What is this function for? Can you elaborate on why other permissions are not compatible for origination? And why do you restrict receiver and sender hooks? I guess there should be some documentation.

sras: There is some documentation here [1]. There is some discussion about this here as well [2]

But i ll add something here too.

[1] https://github.com/tqtezos/stablecoin/pull/34/files#diff-247038a0d8ab37be5403fa80adf5390aR130
[2] https://gitlab.com/morley-framework/morley/-/merge_requests/363#note_336764521

haskell/src/Lorentz/Contracts/Stablecoin.hs
sras: I have removed the type alias now. And instead just case the contract to FA2 parameter in when passing the originated contract to the FA2 tests. I have added the comment at that location.

haskell/src/Lorentz/Contracts/Stablecoin.hs
sras: You are right. Removed them.

haskell/test/Stablecoin.hs
sras: Have fixed.

haskell/test/Stablecoin.hs
sras: Added.

haskell/test/Nettest.hs
sras: Fixed.

awkure:
Kinda fucked up rebase in with ligo files, but fixups are going to be squashed anyway, sorry

PR #36, opened 2020-05-13
Update the specification

### LINKED ISSUE #32: Specification: using term "owner"
https://github.com/tqtezos/stablecoin/blame/master/docs/specification.md#L70
points out that FA2 and CENTRE standards use term "owner" differently.
To avoid confusion I suggest to use qualified terms: "token owner" vs "contract owner"

### LINKED ISSUE #33: Specification: discrepancy in the parameter type for configure_minter and get_minting_allowance
https://github.com/tqtezos/stablecoin/blame/master/docs/specification.md#L609-L626

`current_minting_allowance` field type is `option nat`, but corresponding Michelson has it as `nat` type. 

Ditto for `get_minting_allowance` entry point, `get_minting_allowance_response.alowance` field:
https://github.com/tqtezos/stablecoin/blame/master/docs/specification.md#L650-L679

gromakovsky:
## Description

Another portion of specification updates based on recent discussions.

[Rendered specification](https://github.com/tqtezos/stablecoin/blob/gromak/update-spec-again/docs/specification.md)

## Related issue(s)

Resolves #32 
Resolves #33 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -26,79 +26,6 @@ These specifications were assembled with the following references:
 - The token contract should reject transfers with non-zero XTZ AMOUNT (when someone calls the contract with non-zero amount).
 The reason is that there is no way to spend XTZ owned by the contract.
 
-# FA2 Specifics
-
-FA2 provides a framework for defining permission policies, but does not require any particular policy.
-Moreover, it suggests an approach where permission logic is defined in a different contract and the token contract can dynamically change the policy.
-For the Stablecoin project we need to answer several questions:
-
-1. FA2 recommends the "transfer hook" design pattern, as opposed to implementing a monolithic contract.
-Are there any requirements whether the stablecoin contract should be monolithic or follow this pattern (i. e. have `set_transfer_hook`)?
-2. For a monolithic contract: which permission policy should we implement?
-Should the whole policy or at least some part of it be hardcoded?
-E. g. can we assume that self transfer will always be permitted and there will be no custom policy?
-What about other policies?
-3. For a modular (non-monolithic) contract: what exactly should be implemented by us and put into the repository?
-Should we implement any particular `transfer_hook`?
-If we should, the questions from (2) apply here.
-4. Should we implement any `owner_hook`?
-5. The FA2 standard does not specify who is permitted to update operators on behalf of the token owner.
-Who should be allowed to do it in this project?
-
-Overall, the response from TQ was that there are no strict requirements and we should pick the best approaches from the business logic.
-At the early development stage we propose the following:
-1. We will implement both options (starting from the monolithic one because it's simpler), measure gas costs and then will decide what is better.
-The "transfer hook" approach is recommended, but may be infeasible due to gas costs.
-2. We will implement a single permission policy (as part of the token contract or as a separate transfer_hook – that depends on the previous point).
- + Self transfer is permitted as well as operator transfer.
- + Owner transfer policy for both sender and receiver is `Optional_owner_hook`.
- + No custom permission policy.
-3. We will not implement any `owner_hook`.
-The contract is usable without them because we stick to `Optional_owner_hook`.
-
-Regarding `update_operators` we will stick to the following logic:
-1. In each `update_operator` item `owner` MUST be equal to `SENDER`.
-2. Each address can have arbitrary number of operators.
-
-One more uncertainty is related to the `token_metadata` entrypoint: whether metadata is constant or can be changed during contract's lifetime.
-We assume that token metadata stays constant forever.
-
-Note: these decisions may change as the project moves forward.
-
-# Questions
-
-1. How many addresses can be assigned to each role?
-We have made some assumptions, see the [Roles](#roles) section below, please check them.
-For non-unique roles: is there an upper limit on the number of addresses?
-For unique roles: can there be no address with a certain role (e. g. no pauser)?
-2. Can one address have more than one role?
-3. Are any getter entrypoints not present in FA2 required?
-Should they be `void` or `view`?
-Note that the contract will have storage annotations and each getter entrypoint increases contract's size and gas costs.
-At this point it's hard to predict whether we will have issues with the contract's size and gas costs since we've never used LIGO, but it's quite likely that we will.
-
-# Differences with CENTRE
-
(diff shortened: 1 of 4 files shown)


docs/specification.md
```diff
@@ -748,79 +658,132 @@ Parameter (in Michelson): `list nat`.
...
+3. We will not implement any `owner_hook`.
+The contract is usable without them because we stick to `Optional_owner_hook`.
 
-**set_safelist**
+Regarding `update_operators` we will stick to the following logic:
+1. In each `update_operator` item `owner` MUST be equal to `SENDER`.
+2. Each address can have arbitrary number of operators.
 
-Parameter (in Michelson): `option address`.
+One more uncertainty is related to the `token_metadata` entrypoint: whether metadata is constant or can be changed during contract's lifetime.
```
sras, line 762: If Token_metadata can be changed during the lifetime of the contract, enabling the contract to add support for more token_ids in course of its life, should an operator added using `All_tokens` get automatic access to any new tokens added during the contract's life?

haskell/README.md
```diff
@@ -6,3 +6,14 @@ SPDX-License-Identifier: MIT
...
+
+## Build Instructions
+
+You need [Stack](http://haskellstack.org/) to build this package.
+Run `stack build` to build the library.
+
+## Tests
+
+Tests require the stablecoin contract to be in `test/resources/stablecoin.tz`.
+After you put it there you can do `stack test` to run tests.
```
sras, line 18: `stack test` should be `make test` right? because `stack test` does not build the contract?

gromakovsky: But I am saying "After you put it there". After you put it there you can indeed just run `stack test`. And the next line says that there is a `Makefile` that automates this process. I can add that one can run `make test`.

docs/specification.md
gromakovsky: I am not saying about enabling more token_ids. We will always have only `token_id = 0`. Each token ID has its own metadata and I am talking only about metadata for `token_id = 0`. Do we want to change things like `name` or `extras` after origination? We assume that no.

docs/specification.md
```diff
@@ -748,79 +658,132 @@ Parameter (in Michelson): `list nat`.
...
-## Safelist update
+Overall, the response from TQ was that there are no strict requirements and we should pick the best approaches from the business logic.
+At the early development stage we propose the following:
+1. We will implement both options (starting from the monolithic one because it's simpler), measure gas costs and then will decide what is better.
+The "transfer hook" approach is recommended, but may be infeasible due to gas costs.
+2. We will implement a single permission policy (as part of the token contract or as a separate transfer_hook – that depends on the previous point).
+ + Self transfer is permitted as well as operator transfer.
+ + Owner transfer policy for both sender and receiver is `Optional_owner_hook`.
+ + No custom permission policy.
+3. We will not implement any `owner_hook`.
```
awkure, line 755: Missed an actual third answer

docs/specification.md
```diff
@@ -748,79 +658,132 @@ Parameter (in Michelson): `list nat`.
...
+The contract is usable without them because we stick to `Optional_owner_hook`.
 
-**set_safelist**
+Regarding `update_operators` we will stick to the following logic:
+1. In each `update_operator` item `owner` MUST be equal to `SENDER`.
+2. Each address can have arbitrary number of operators.
```
awkure, line 760: Also we need to clarify their non-transitive behavior as was discussed in https://gitlab.com/tzip/tzip/-/issues/17

ligo/README.md
```diff
@@ -3,3 +3,10 @@ SPDX-FileCopyrightText: 2020 tqtezos
 SPDX-License-Identifier: MIT
 -->
 # stablecoin LIGO implementation
+
+Implementation of a token smart contract in LIGO according to the [specification](/docs/specification.md).
+
+## Build Instructions
+
+1. Make sure you have `ligo` executable in your `$PATH` or set `LIGO` environment variable to the command you use to invoke `ligo`.
+2. Run `make`, it should produce `stablecoin.tz`.
```
awkure, line 12: I think it's better to clarify its exact path

gromakovsky: But `stablecoin.tz` is the exact path, isn't it? If you run `make`, it will produce `stablecoin.tz` in the current folder.

docs/specification.md
gromakovsky: What do you mean? I only moved this text from top to bottom. If you mean the question (3) above, I think it's answered by (1) and (2). A modular contract will have the same functionality as the monolithic one, but will be modular.

ligo/README.md
awkure: Ah, I thought that we agreed to move generated contract to `haskell/test/resources` as was done in one of our PRs, ok then.

gromakovsky: `make` in `ligo` produces `stablecoin.tz` (in current directory). If you call `make test` from the root or from `haskell/`, it will build `stablecoin.tz` and then symlink it to `haskell/test/resources`.

docs/specification.md
gromakovsky: Added a line.

docs/specification.md
```diff
@@ -490,19 +441,18 @@ Parameter (in Michelson)
 - This entrypoint MUST follow the FA2 requirements.
 
 - Permission logic requirements specific to this contract are described in the ["FA2 Specifics"](#fa2-specifics) chapter.
-Also see [this issue](https://gitlab.com/tzip/tzip/-/issues/16).
+Each `owner` must be equal to `SENDER`.
 
 - Since the contract supports only a single token type, all `token_id` values MUST be 0.
   They are passed because FA2 requires that.
-  So there are at most three possible valid values of `operator_tokens` type:
+  So there are three possible valid values of `operator_tokens` type:
```
emishur, line 448: Check recent updates to FA2 standard. Operators are simplified and do not have per token granularity anymore

gromakovsky:
@emishur at this point we are a bit off the schedule and I would like to stabilize things as they are. If we keep trying to follow FA2 we will likely not merge anything for a few more days and there is already a long delay. I am going to merge this PR, but I have created a new issue to update the specification (and code) according to the recent FA2 changes: #39. We will do it once we have first (corresponding to a bit outdated FA2) version of the contract.

PR #38, opened 2020-05-14
[#31] Implement management entrypoints

### LINKED ISSUE #31: Implement management entrypoints

After #8 we should have FA2 entrypoints, but there are other entrypoints are contract needs.

# Acceptance criteria

The following entrypoints are implemented according to the specification:
* `pause`
* `unpause`
* `get_paused`
* `configure_minter`
* `remove_minter`
* `get_minting_allowance`
* `mint`
* `burn`
* `transfer_ownership`
* `get_owner`
* `accept_ownership`
* `change_master_minter`
* `get_master_minter`
* `change_pauser`
* `get_pauser`

There are tests (including network tests) for them.

Safelist stuff will be done later.

awkure:
## Description

This MR implements the rest of required entrypoints

- [X] pause
- [X] unpause 
- [X] configure_minter
- [X] remove_minter
- [x] mint
- [x] burn
- [X] transfer_ownership
- [X] accept_ownership
- [X] change_master_minter
- [X] change_pauser

## Related issue(s)

Resolves #31 #39 
partly #53 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


ligo/stablecoin/actions.ligo
```diff
@@ -13,13 +13,101 @@
...
+function authorize_master_minter
+  ( const store : storage
+  ) : unit is
+  fail_on
+    ( Tezos.sender =/= store.roles.master_minter
+    , "Sender is not master minter"
+    )
+
+(*
+ * Authorizes master_minter and fails otherwise.
```
sras, line 73: Probably should be `minter` instead of `master_minter` here.

ligo/stablecoin/actions.ligo
```diff
@@ -30,16 +118,18 @@ function credit_to
 ; updated_balance := updated_balance + parameter.amount
 ; const updated_ledger : ledger =
     Big_map.update(parameter.to_, Some (updated_balance), store.ledger)
-; const ups : storage = store with record [ ledger = updated_ledger ]
-} with ups
+} with store with record
+    [ ledger = updated_ledger
+    ; total_supply = store.total_supply + parameter.amount
```
sras, line 123: Why is total_supply is being incremented here?

ligo/stablecoin/actions.ligo
```diff
@@ -52,7 +142,10 @@ function debit_from
       }
   | None -> failwith ("Given address is not present in ledger")
   end
-} with store with record [ ledger = updated_ledger ]
+} with store with record
+    [ ledger = updated_ledger
+    ; total_supply = abs (store.total_supply - parameter.amount) // TODO: possible bug here
```
sras, line 147: Same here. Why is `total_supply` touched here?

ligo/stablecoin/actions.ligo
awkure: It's used as a slight workaround which helps us not bothering about total supply updates when working with tokens

ligo/stablecoin/actions.ligo
```diff
@@ -13,13 +13,101 @@
...
+ *)
+function authorize_minter
+  ( const store : storage
+  ) : unit is case store.minting_allowances[Tezos.sender] of
+    Some (u) -> unit
+  | None -> failwith ("Sender is not minter")
+  end
+
+(*
+ * Helper that fails if the given contract is paused.
```
sras, line 83: 'given contract' ?

ligo/stablecoin/actions.ligo
```diff
@@ -13,13 +13,101 @@
...
+function ensure_not_paused
+  ( const store : storage
+  ) : unit is
+  fail_on
+    ( store.paused
+    , "Given operation is restricted during contract pause"
+    )
+
+(*
+ * Helper that fails if the given contract is paused.
```
sras, line 94: Should be 'is not paused' in think...

ligo/stablecoin/actions.ligo
sras: I don't get it. Can you please explain? What issue does this work around solve?

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +339,230 @@ function is_operator_action
...
+    None -> case store.minting_allowances[parameter.0] of
+      Some (u) -> failwith ("Minter already present in minter list which was not expected")
+    | None -> skip
+    end
+  | Some (current_minting_allowance) ->
+      case store.minting_allowances[parameter.0] of
+        Some (allowance) -> if allowance =/= current_minting_allowance
+          then failwith ("Current and expected minting allowances do not match")
+          else skip
+      | None -> failwith ("Minter is not present in minter list althought is expected")
```
sras, line 398: >althought

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +339,230 @@ function is_operator_action
...
+ *)
+function configure_minter
+  ( const parameter : configure_minter_params
+  ; const store     : storage
+  ) : entrypoint is block
+{ ensure_not_paused (store)
+; authorize_master_minter (store)
+; case parameter.1.0 of
+    None -> case store.minting_allowances[parameter.0] of
+      Some (u) -> failwith ("Minter already present in minter list which was not expected")
```
sras, line 390: Error message could say something like current minting allowance is required to change it to new value.

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +339,230 @@ function is_operator_action
...
+          | Some (current_allowance) ->
+              mint_tokens (accumulator, mint_param, current_allowance)
+          end
+      , parameters
+      , store
+      )
+  )
+
+(*
+ * Decreases balance and total supply of otkens by the given amount
```
sras, line 484: > otkens

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +339,230 @@ function is_operator_action
...
+                ]
+            , accumulator
+            )
+      , parameters
+      , store
+      )
+  )
+
+(*
+ * Sets token owner to a new address
```
sras, line 512: I think it should be "contract owner" here.

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +339,230 @@ function is_operator_action
...
+  ( const parameter : accept_ownership_param
+  ; const store     : storage
+  ) : entrypoint is block
+{ authorize_pending_owner (store)
+} with
+  ( (nil : list (operation))
+  , store with record
+      [ roles.pending_owner = (None : option (address))
+      // We have already validated pending owner
+      ; roles.owner = Tezos.sender
```
sras, line 538: It might be better to be defensive and explicitly use the value in the `pending_owner` field here, instead of using the sender value.

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +339,230 @@ function is_operator_action
...
+      ]
+  )
+
+(*
+ * Moves master minter priveleges to a new address.
+ * Fails if sender is not master minter.
```
sras, line 544: >  * Fails if sender is not master minter.

is not owner, right?

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +339,230 @@ function is_operator_action
...
+  ) : entrypoint is block
+{ authorize_owner (store)
+} with
+  ( (nil : list (operation))
+  , store with record [ roles.master_minter = parameter ]
+  )
+
+(*
+ * Moves pauser priveleges to a new address.
+ * Fails if sender is not current pauser.
```
sras, line 558: >  * Fails if sender is not current pauser.

think should be "is not owner" here as well..

ligo/stablecoin/actions.ligo
awkure: I wonder if it requires more instructions

ligo/stablecoin/actions.ligo
awkure: Ok, I changed it so total supply can be modified only from `mint` and `burn` entrypoints

haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
```diff
@@ -17,7 +17,8 @@ module Lorentz.Contracts.Spec.FA2Interface
   , OperatorTransferMode (..)
   , OwnerTransferMode (..)
   , Parameter (..)
-  , ParameterC
+  , FA2ParameterC
+  , FA2ParameterMichelsonEntrypoints
```
sras, line 21: We are having the FA2 stuff here as a temporary arrangement, and we will eventually import all FA2 interface stuff/tests from Morley. So may be it is better not to depend on specific changes to this module.

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -0,0 +1,504 @@
...
+      withOriginated originate defaultOriginationParams $ \stablecoinContract -> do
+        lCallEP stablecoinContract (Call @"Pause") ()
+        lCallEP stablecoinContract (Call @"Pause") ()
+        validate . Left $ lExpectAnyMichelsonFailed stablecoinContract
+
+    it "unpauses contract as expected" $ integrationalTestExpectation $ do
+      let originationParams = defaultOriginationParams { opPaused = True }
+      withOriginated originate originationParams $ \stablecoinContract -> do
+        lCallEP stablecoinContract (Call @"Unpause") ()
+        validate . Right $ expectAnySuccess
```
sras, line 45: Shouldn't we also check the pause status was changed somehow?

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -0,0 +1,504 @@
...
+
+    it "prevents transfers while contract is paused" $ integrationalTestExpectation $ do
+      let
+        originationParams =
+            addAccount (wallet1, (commonOperators, 10))
+          $ addAccount (wallet2, ([], 0))
+          $ defaultOriginationParams
+              { opPermissionsDescriptor = permissionDescriptorOperatorTransferAllowed }
+      withOriginated originate originationParams $ \stablecoinContract -> do
+        lCallEP stablecoinContract (Call @"Pause") ()
```
sras, line 67: Why not set the pause status while originating?

haskell/test/Lorentz/Contracts/Test/Management.hs
sras: It seems that we have another test that check the above. So doesn't that make this test redundant?

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -0,0 +1,504 @@
...
+    it "configures minter properly" $ integrationalTestExpectation $ do
+      withOriginated originate defaultOriginationParams $ \stablecoinContract -> do
+        let
+          configureMinterParam1 =
+            ( #minter .! wallet2
+            , ( #current_minting_allowance .! Nothing
+              , #new_minting_allowance .! 20
+              ))
+
+        withSender admin $ lCallEP stablecoinContract (Call @"Configure_minter") configureMinterParam1
```
sras, line 129: Why is admin calling this entrypoint. Isn't this supposed to be called by master minter?

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -0,0 +1,504 @@
...
+            , (#request .! (#owner .! wallet2, #token_id .! 0), #balance .! 5)
+            , (#request .! (#owner .! wallet3, #token_id .! 0), #balance .! 15)
+            ]
+
+        lCallEP stablecoinContract (Call @"Balance_of") balanceRequest
+
+        validate . Right $
+          lExpectViewConsumerStorage consumer [balanceExpected]
+
+    it "aborts whole transaction if one of the amount of minting tokens exceeds current minting allowance" $ integrationalTestExpectation $ do
```
sras, line 292: Why 'one of the amount', and not the sum of all?

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -0,0 +1,504 @@
...
+
+
+  describe "Contract ownership" $ do
+    it "transfers ownership properly" $ integrationalTestExpectation $ do
+      withOriginated originate defaultOriginationParams $ \stablecoinContract -> do
+        withSender admin $ lCallEP stablecoinContract (Call @"Transfer_ownership") wallet1
+        withSender wallet1 $ lCallEP stablecoinContract (Call @"Accept_ownership") ()
+        withSender wallet1 $ lCallEP stablecoinContract (Call @"Transfer_ownership") wallet2
+        withSender wallet2 $ lCallEP stablecoinContract (Call @"Accept_ownership") ()
+        validate . Right $ expectAnySuccess
```
sras, line 423: Shouldn't we check the ownership switch actually happned?

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -0,0 +1,504 @@
...
+        withSender admin $ lCallEP stablecoinContract (Call @"Transfer_ownership") wallet1
+        withSender admin $ lCallEP stablecoinContract (Call @"Transfer_ownership") wallet2
+        withSender wallet1 $ lCallEP stablecoinContract (Call @"Accept_ownership") ()
+        validate . Left $ lExpectAnyMichelsonFailed stablecoinContract
+
+
+
+  describe "Master minter" $ do
+    it "changes contract master minter properly" $ integrationalTestExpectation $ do
+      withOriginated originate defaultOriginationParams $ \stablecoinContract -> do
```
sras, line 468: I think origination function should accept admin, pauser and master minter addresses for these tests to really make sense.

haskell/test/Lorentz/Contracts/Test/Management.hs
awkure: Done

haskell/test/Lorentz/Contracts/Test/Management.hs
awkure: Changed

haskell/test/Lorentz/Contracts/Test/Management.hs
awkure: We have no `admin` role in stablecoin contract. Here admin is an actual superuser that has all roles assigned to him. Regarding to permissions validation there's already present test (`fails if sender does not have master minter permissions`) for it.

haskell/test/Lorentz/Contracts/Test/Management.hs
awkure: Changed

haskell/test/Lorentz/Contracts/Test/Management.hs
awkure: The point I was trying to make is to simplify things as possible by making only one superuser. I've changed its name if this clarifies some things. Also, I've added the capability of changing roles during origination, but it still seems to be unused.

haskell/test/Lorentz/Contracts/Test/Management.hs
sras: > Here admin is an actual superuser that has all roles assigned to him. 

Yes, I was referring to this. Basically, I think we should have different addresses for owner, master minter and pause roles. Or else this won't catch if the owner actually allows the owner (I mean the admin here), to call the entrypoints that are only callable by other two roles, and vice versa.

haskell/test/Lorentz/Contracts/Test/Management.hs
sras: I am not sure if we can test that the roles work well if you assign all roles to a single address.

>but it still seems to be unused.

Didn't get you.

gromakovsky:
I will review after @sras approves it.

haskell/test/Lorentz/Contracts/Test/Management.hs
sras: I cannot seem to find the corresponding change. What was it?

ligo/stablecoin/actions.ligo
sras: @awkure The comments are still mixed up I think. For 'ensure_is_paused' it says 'fails if contract is paused' and for 'ensure_not_paused' it says 'fails if contract is not paused.'

Also the message for `ensure_is_paused` reads "Contract is paused", when it should be something like "contract should be paused, but isn't".

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +333,232 @@ function is_operator_action
...
+    ) : storage is block
+  { const debited_storage : storage = debit_from
+      ( record
+          [ from_ = sender_address
+          ; amount = burn_amount
+          ]
+      , accumulator
+      )
+  } with debited_storage with record
+      [ total_supply = abs (debited_storage.total_supply - burn_amount) ]
```
sras, line 501: Why is `abs` required here?

awkure: This error won't be called anyway since in that case `debit_from` would fail earlier, but anyway I've changed it

sras: Yes, of course. We woulndn't encounter that case normally, unless there is some bug else where. But if there is a bug else where and we indeed hit this case, having `abs` function here will result in silently corrupted contract state, right, that is of course unless our tests don't catch it. 

But I think it is better not to even leave that possibility in code..

haskell/test/Lorentz/Contracts/Test/Management.hs
awkure: I mean that previously we could not change originated params and had set up default permissions during origination. While testing we still change roles by calling specific entrypoints and validating it after by using specific patterns. It's not essential tho, just mentioning.

ligo/stablecoin/actions.ligo
sras: @awkure This still needs to be fixed from the earlier comment.

>Also the message for ensure_is_paused reads "Contract is paused", when it should be something like "contract should be paused, but isn't".

awkure: updated

haskell/test/Stablecoin.hs
```diff
@@ -26,11 +29,23 @@ import Tezos.Core
 contractPath :: FilePath
 contractPath = "test/resources/stablecoin.tz"
 
-spec_Stablecoin :: Spec
-spec_Stablecoin = specWithUntypedContract contractPath specCallback
+origination :: forall param . U.Contract -> OriginationFn param
+origination contract (mkInitialStorage -> Just storageVal) =
+  -- We just cast the stablecoin contract address as an FA2 contract
```
gromakovsky, line 34: This comment looks wrong now. We cast to abstract `param` now, not to FA2 parameter.

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +333,236 @@ function is_operator_action
...
+
+(*
+ * Moves master minter priveleges to a new address.
+ * Fails if sender is not contract owner.
+ *)
+function change_master_minter
+  ( const parameter : change_master_minter_param
+  ; const store     : storage
+  ) : entrypoint is block
+{ authorize_master_minter (store)
```
gromakovsky, line 550: It looks wrong. It will check that the sender is the current master minter, but it should check that the sender is the current contract owner. Same for `change_pauser`.

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +333,236 @@ function is_operator_action
...
+ *)
+function configure_minter
+  ( const parameter : configure_minter_params
+  ; const store     : storage
+  ) : entrypoint is block
+{ ensure_not_paused (store)
+; authorize_master_minter (store)
+; case parameter.1.0 of
+    None -> case store.minting_allowances[parameter.0] of
+      Some (u) -> failwith ("Current minting allowance is required for changing already set minting allowance for the given address")
```
gromakovsky, line 384: Please use shorter error messages. Each character here adds 1 byte to the contract size and increases gas costs. Usually descriptive error messages are good, but in case of Michelson they may cause troubles. So I think we should prefer brevity to descriptiveness. This one is particularly large and definitely should be made shorter. Other messages are also quite large and I would make them shorter as well, but that's up to you.

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +333,236 @@ function is_operator_action
...
+} with
+  ( (nil : list (operation))
+  , store with record [paused = False]
+  )
+
+(*
+ * Adds minter to minting allowances map setting its allowance
+ * to 0. Resets the allowance if associated address is already
+ * present. Fails if sender is not master minter.
+ * TODO
```
gromakovsky, line 374: TODO what?

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +333,236 @@ function is_operator_action
...
+  ) : entrypoint is block
+{ ensure_not_paused (store)
+; authorize_master_minter (store)
+; case parameter.1.0 of
+    None -> case store.minting_allowances[parameter.0] of
+      Some (u) -> failwith ("Current minting allowance is required for changing already set minting allowance for the given address")
+    | None -> skip
+    end
+  | Some (current_minting_allowance) ->
+      case store.minting_allowances[parameter.0] of
```
gromakovsky, line 388: Would it be beneficial to define a variable/constant holding `store.minting_allowances[parameter.0]`? It takes some non-trivial computation to obtain this value. If we define `cur_allowance = store.minting_allowances[parameter.0]` will it be computed once?

ligo/stablecoin/actions.ligo
gromakovsky: Apparently tests are also broken.

sras: Curious. It seems that it was `authorize_owner` for both `change_pauser` and `change_master_minter` entrypoints in the commit that I reviewed.

https://github.com/tqtezos/stablecoin/blob/2866d4fa4884ebd1c0de85c856dda5109a33c809/ligo/stablecoin/actions.ligo#L550

gromakovsky:
> Please make sure the contract can be originated in carthagenet.

Okay, now it will be tested automatically by CI.

ligo/stablecoin/actions.ligo
awkure: Implemented already, forgot to remove comment

haskell/test/Stablecoin.hs
gromakovsky: This comment still looks wrong.

ligo/stablecoin/actions.ligo
```diff
@@ -13,13 +13,101 @@
...
+
+(*
+ * Ensures that sender is current pending owner
+ *)
+function authorize_pending_owner
+  ( const store : storage
+  ) : unit is case store.roles.pending_owner of
+    Some (pending_owner) -> fail_on
+      ( Tezos.sender =/= pending_owner
+      , "Sender is not current pending owner"
```
sras, line 45: Why the difference in fail message formats? ie "NOT_PENDING_OWNER" and "Sender is not current pending owner" ?
Also shouldn't the former say "Contract not in transfer" or something like that?

ligo/stablecoin/actions.ligo
```diff
@@ -13,13 +13,101 @@
...
+      )
+    | None -> failwith ("NOT_PENDING_OWNER")
+    end
+
+(*
+ * Authorizes paused and fails otherwise.
```
sras, line 51: > Authorizes paused

Should probably be "pauser"

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +338,237 @@ function is_operator_action
...
+      [ minting_allowances = Big_map.update
+          ( parameter.0
+          , Some (parameter.1.1)
+          , store.minting_allowances
+          )
+      ]
+  )
+
+(*
+ * Removes minter from minting allowances map. Does nothing if
```
sras, line 413: >Does nothing 

AFAIU, it fails with "NOT_MINTER"

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +338,237 @@ function is_operator_action
...
+    ) : storage is block
+  { const unwrapped_parameter : mint_param_ =
+      Layout.convert_from_right_comb ((mint_param : mint_param))
+  ; fail_on
+      ( unwrapped_parameter.amount > current_allowance
+      , "ALLOWANCE_EXCEEDED"
+      )
+
+  ; const updated_allowance : nat =
+      abs (current_allowance - unwrapped_parameter.amount)
```
sras, line 460: Might be better if we combine the `fail_on` call above above and this assignment using the `IS_NAT` check.

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +338,237 @@ function is_operator_action
...
+      ( record
+          [ from_ = sender_address
+          ; amount = burn_amount
+          ]
+      , accumulator
+      )
+  } with debited_storage with record
+      [ total_supply = case is_nat (debited_storage.total_supply - burn_amount) of
+          Some (nat) -> nat
+        | None -> (failwith ("NEGATIVE_TOTAL_SUPPLY") : nat)
```
sras, line 509: May be we should mark this as a contract error somehow, that should never happen in normal operation.

ligo/stablecoin/actions.ligo
sras: On second look, I fear this appears kind of important. Because if by some accident, the `authorize_pending_owner` call is removed or if there is a bug in it (a bad rebase, for example), it will give any random sender owner privileges. @gromakovsky What do you think?

I think `authorize_pending_owner` can be made to return the address of pending owner as `address` (unwrapping it from `Some` value or failing otherwise), and we can just use it here. Shouldn't add a lot more instructions, right?

ligo/stablecoin/actions.ligo
awkure: Oh, have missed it, thanks

ligo/stablecoin/actions.ligo
awkure: We can clarify these errors in our spec 🤔

haskell/test-common/Lorentz/Contracts/Test/Management.hs
```diff
@@ -0,0 +1,557 @@
+-- SPDX-FileCopyrightText: 2020 tqtezos
+-- SPDX-License-Identifier: MIT
+
+-- | Tests for management entrypoints of stablecoin smart-contract
+
+module Lorentz.Contracts.Test.Management
```
gromakovsky, line 6: Why is this module in `test-common`?

ligo/stablecoin/actions.ligo
gromakovsky: Let's do it separately, see https://github.com/tqtezos/stablecoin/issues/53

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +338,237 @@ function is_operator_action
...
+  ; const store     : storage
+  ) : entrypoint is block
+{ authorize_owner (store)
+} with
+  ( (nil : list (operation))
+  , store with record [roles.pending_owner = Some (parameter)]
+  )
+
+(*
+ * When pending sender calls this entrypoint, the address
```
gromakovsky, line 531:
```suggestion
 * When pending contract owner calls this entrypoint, the address
```

ligo/stablecoin/actions.ligo
```diff
@@ -240,4 +338,237 @@ function is_operator_action
...
+ * receives owner privileges for contract
+ *)
+function accept_ownership
+  ( const parameter : accept_ownership_param
```
gromakovsky, line 535: `accept_ownership_param` is `unit`. Isn't it more efficient to not pass it here at all?

ligo/stablecoin/actions.ligo
gromakovsky: > I think authorize_pending_owner can be made to return the address of pending owner as `address`

Okay, sounds cheap indeed, let's do it.

haskell/test-common/Lorentz/Contracts/Test/Management.hs
awkure: Because `FA2.hs` was 🤷

ligo/stablecoin/actions.ligo
awkure: It's stripped anyway 🤷

haskell/test-common/Lorentz/Contracts/Test/Management.hs
gromakovsky: `FA2.hs` is used by both test-suites, `Managemenet.hs` is not.

awkure: But it will be once I start working on nettest scenarios anyway

ligo/stablecoin/actions.ligo
awkure: done

ligo/stablecoin/actions.ligo
awkure: done

ligo/stablecoin/actions.ligo
awkure: done

ligo/stablecoin/actions.ligo
awkure: done

ligo/stablecoin/actions.ligo
gromakovsky: So will this function have type `lambda storage foo`? Not `lambda (pair unit storage) foo` (for some `foo`)? Did you check that?

haskell/test-common/Lorentz/Contracts/Test/Management.hs
gromakovsky: Why do you think so? It has a large portion of `hspec` stuff. How do you plan to use it in nettest?

gromakovsky: That being said, `FA2.hs` is not used by nettest anymore either, so I guess it's time to move it to `test/` (as well as `Management.hs`).

ligo/stablecoin/actions.ligo
awkure: Yes, it will be inlined

ligo/stablecoin/spec.ligo
```diff
@@ -173,18 +179,41 @@ type permissions_descriptor is michelson_pair_right_comb(permissions_descriptor_
...
+
+type accept_ownership_param is unit
+
+type change_master_minter_param is address
+
+type change_pauser_param is address
+
 type parameter is
-  Transfer               of transfer_params
+  Transfer               of transfer_param
```
sras, line 216: @awkure As per FA2 parameter of transfer is still `list(transfer_param)` right? So shouldn't it be `transfer_params` here ?

awkure: No, the list is now enclosed in `parameter.destinations`

gromakovsky: This is what we have now:
```
                               (pair %transfer
                                  (address %from_)
                                  (list %destinations (pair (address %to_) (pair (nat %token_id) (nat %amount))))))
```

Compare it with FA2, it's different: https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-12/tzip-12.md#transfer

gromakovsky: Also `%destinations` should be changed to `%txs`.

PR #46, opened 2020-05-20
[#37] Nettest local chain in ci

### LINKED ISSUE #37: Run nettest scenarios on CI

Running nettest scenarios requires some setup, e. g. `tezos-client` executable ought to be available. We have solved such a task in `morley`, see https://gitlab.com/morley-framework/morley/-/issues/203 and https://gitlab.com/morley-framework/morley/-/merge_requests/408. Now we should do it here as well.

Note: currently we have no nettest scenarios, we should merge #34 first, so please don't start this issue before that. Besides, we probably need https://gitlab.com/morley-framework/morley/-/merge_requests/407.

# Acceptance criteria

1. In each PR CI runs nettests for stablecoin on local chain.
2. If it's not hard to do, it should also run it on testnet (node address can be hardcoded and updated manually) regularly (not in each PR but time-based).

rvem:
## Description

Added nettest scenario on real chain run to the CI.

~However, scenario currently fails with an error somewhere inside the `tezos-baker-lib`:(~

## Related issue(s)

Resolves #37

- Tests
        silently reappearing again.

- Documentation
    - Haddock


haskell/test-common/Lorentz/Contracts/Test/Common.hs
```diff
@@ -0,0 +1,67 @@
+-- SPDX-FileCopyrightText: 2020 tqtezos
+-- SPDX-License-Identifier: MIT
+module Lorentz.Contracts.Test.Common
```
gromakovsky, line 3: If we run nettest test-suite on CI, we don't need to include this scenario into two test-suites, it's sufficient to put it into only one (like we did in `morley-ledgers-test`).

haskell/Makefile
```diff
@@ -45,6 +49,13 @@ test-dumb-term:
 test-hide-successes:
 	TERM=dumb $(call call_test,"--hide-successes",$(STACK_DEV_TEST_OPTIONS))
 
+# Run network tests
+nettest:
```
gromakovsky, line 53: You should add it to the root `Makefile` as well btw.

haskell/test-common/Lorentz/Contracts/Test/Common.hs
gromakovsky: Well, maybe something should still be in test-common, but you definitely put much more than necessary. And `scTransferScenario` shouldn't be there.

gromakovsky:
Btw, I think `scTransferScenario` should operate with concrete `Storage`, but currently it's abstract `storage`.

haskell/Makefile
```diff
@@ -20,11 +20,15 @@ PACKAGE = stablecoin
 #   Initial `../..` in `ln` is needed because of `test/resources`.
 #   $(LIGO_DIR) is relative to the current directory.
 define call_test
+	$(call build_stablecoin_contract,"","")
```
gromakovsky, line 23: AFAIU `"",""` things here are like functions arguments. You can use `$1` and `$2` there. But apparently you don't use them. So why do you pass them?

rvem: copypaste is evil

gromakovsky:
> Problem: `tezos-baker-lib` fails with an error when trying to serialize `LAMBDA` instruction.

Please refer to https://gitlab.com/obsidian.systems/tezos-bake-monitor-lib/-/issues/1. Though maybe we will just drop that commit because it's fixed upstream already and maybe they will merge it quickly :)

PR #49, opened 2020-05-21
[#44] Forget about the transfer hook

### LINKED ISSUE #44: Implement the contract with transfer hook

FA2 suggests a modular contract design with `set_transfer_hook` entrypoint and a transfer hook contract. It's a bit more complicated, but modularity is good. We should implement it and compare it with the existing (monolithic approach).

# Acceptance criteria

1. `stablecoin` contract is implemented using transfer hook design.
2. We measure costs of its operations.
3. Then we decide whether we want to replace the existing version with the modular one or maybe even keep both.

awkure:
## Description

See https://github.com/tqtezos/stablecoin/pull/49#issuecomment-641220397

## Related issue(s)

Resolves #44 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -751,9 +751,10 @@ If we should, the questions from (2) apply here.
 Who should be allowed to do it in this project?
 
 Overall, the response from TQ was that there are no strict requirements and we should pick the best approaches from the business logic.
-At the early development stage we propose the following:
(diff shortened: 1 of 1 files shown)


gromakovsky:
We heard that:
1. Transfer hook pattern implementation will likely fail to fit into gas limits.
2. Transfer hook will likely stop being recommended pattern because Tezos is not mature enough for inter-contract operations.

For this reason we are not going to continue this effort.

Issue #43, opened 2020-05-20
Add state machine tests for non-FA2 operations


gromakovsky:
# Clarification and motivation

In #42 we are going to add SMT to FA2 operations. But we need to cover other operations as well.

# Acceptance criteria

State machine testing is extended to cover all entrypoints. Note that tests for FA2 entrypoints should be separated because we are going to move them to `morley-ledgers-test` later.

Please don't start this issue before at least #42, #35, #31 are done.

gromakovsky:
Resolved in https://github.com/tqtezos/stablecoin/pull/51

PR #51, opened 2020-05-25
[#42]  SMT tests

### LINKED ISSUE #42: Add state machine tests for FA2 operations

In #13 some tests for FA2 were added, but there are no state machine tests. State machine tests should give us more confidence about correctness of our contract.

# Acceptance criteria

There are state machine tests for FA2 entrypoints similar to those we have for FA1.2 entrypoints [in this package](https://gitlab.com/morley-framework/morley/-/tree/master/code/morley-ledgers-test). CI runs them and they pass.

sras:
## Description

Adds SMT tests for stablecoin contract.

## Related issue(s)

Resolves #42 #43

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/package.yaml
+++ b/haskell/package.yaml
@@ -47,9 +47,12 @@ tests:
       - tasty-hunit-compat
       - tasty-hspec
       - tasty-quickcheck
+      - QuickCheck
       - stablecoin
       - hspec
       - containers
+      - mtl
+      - fmt
 
   stablecoin-nettest:
     <<: *test-common
--- a/haskell/test/SMT.hs
+++ b/haskell/test/SMT.hs
@@ -1,0 +1,664 @@
+-- SPDX-FileCopyrightText: 2020 tqtezos
+-- SPDX-License-Identifier: MIT
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module SMT
+  ( smtProperty
+  ) where
+
+import Data.Maybe (fromJust)
+import qualified Data.Map as Map
+import Data.Typeable (cast)
+import Fmt
+
+import Test.Tasty.QuickCheck
+  ( Arbitrary, Gen, Property, arbitrary, choose, arbitraryBoundedEnum, elements
+  , sublistOf, shuffle, vectorOf)
+import qualified Text.Show
+
+import Tezos.Core (unsafeMkMutez)
+
+import Lorentz (NiceParameter, arg, toVal, fromVal, def)
+import Lorentz.Address
+import Lorentz.Contracts.Stablecoin
+import Lorentz.Contracts.Test.FA2
+import Lorentz.Contracts.Test.Common
+import Michelson.Test.Integrational
+import Michelson.Test.Dummy
+import Michelson.Interpret
+import Michelson.Typed (untypeValue)
+import Michelson.Text
+import Michelson.TypeCheck
+import qualified Michelson.Typed as T
+import qualified Michelson.Untyped as U
+import Util.Named
+
+-- Some implementation notes:
+--
+-- In this module we test the contract by implementing whole contract
+-- functionality in a Haskell, and comparing its behavior against the actual
+-- contract when run using a Michelson interpretr.
+--
+-- First we generate a random series of (sender address, contract parameter)
+-- pairs. Then we generated a random initial contract state.
+-- Then we apply each of these, both to the Haskell model, and on the
+-- michelson interpretation. After each iteration we check the contract state has not
+-- diverged. If it did, we indicate a test failure. If it did not, we extract the new
+-- contract state and initialize both models with it and apply next parameter, and so on.
+--
+-- For random input generation, we first assign random pool of addresses for
+-- each contract roles.  When we make a contract parameter, say transfer, we
+-- fetch a random sender address from the pool of token owners, and to/from
+-- address (for now we pick operator address and token owner address from the
+-- same pool). This is supposed to create inputs that have a little more
+-- chance of being valid input that has a chance to execute successfully.
+
+-- Right now safelist interaction is not tested, since it seems that doing it
+-- in a useful way require modeling a safelist contract as well.
+
+-- | Assign pools of random address for main roles. Then generate
+-- a list of contract calls of given size. When creating the calls, addresses
+-- for various roles are not generated randomly, but randomly choosen from the
+-- pool for the corresponding role. This give us some assurance that some of these
+-- actions will be legal, and will be executed succesfully.
+generateContractInputs :: Int -> Gen ([ContractCall Parameter], ContractState)
+generateContractInputs count = do
+  gsOwnerPool <- vectorOf poolSize (arbitrary @Address)
+  gsMasterMinterPool <- vectorOf poolSize (arbitrary @Address)
+  gsPauserPool <- vectorOf poolSize (arbitrary @Address)
+  gsTokenOwnerPool <- vectorOf poolSize (arbitrary @Address)
+  gsMinterPool <- Map.fromList <$> (fmap unMintingAllowance) <$> (vectorOf 3 (arbitrary @MintingAllowance))
+  owner <- elements gsOwnerPool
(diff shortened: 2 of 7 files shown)


haskell/src/Indigo/Contracts/Safelist.hs.orig
```diff
@@ -0,0 +1,80 @@
...
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+module Indigo.Contracts.Safelist
+  ( Parameter
+  , Storage (..)
+  , mkStorage
+  , safelistContract
+  ) where
+
+<<<<<<< HEAD
```
awkure, line 15: 👀

haskell/src/Lorentz/Contracts/Stablecoin.hs.orig
```diff
@@ -0,0 +1,254 @@
+-- SPDX-FileCopyrightText: 2020 tqtezos
```
awkure, line 1: this module should be deleted

sras: Damn. Fixed.
This won't happen again.

haskell/src/Indigo/Contracts/Safelist.hs.orig
sras: Removed the file.

haskell/package.yaml
```diff
@@ -47,9 +47,13 @@ tests:
       - tasty-hunit-compat
       - tasty-hspec
       - tasty-quickcheck
+      - QuickCheck
       - stablecoin
+      - splitmix
```
gromakovsky, line 52: Is it used?

ligo/stablecoin/operator.ligo
```diff
@@ -13,7 +13,7 @@
 function validate_operator_owner_is_sender
   ( const param : operator_param
   ) : unit is if param.0 = Tezos.sender
-  then unit else failwith ("NOT_OWNER")
+  then unit else failwith ("NOT_TOKEN_OWNER")
```
gromakovsky, line 16: I have cleaned up some errors in #61. Apparently all contract changes in this PR only modify some error messages. If there is a good reason to update some message, let's do it, but also update `specification.md`. If there is no _really good_ reason, let's not change error messages because some contract version has been released already and I think we should make only really necessary changes, not cosmetic ones.

haskell/test/SMT.hs
```diff
@@ -0,0 +1,808 @@
...
+-- each contract roles.  When we make a contract parameter, say transfer, we
+-- fetch a random sender address from the pool of token owners, and to/from
+-- address (for now we pick operator address and token owner address from the
+-- same pool). This is supposed to create inputs that have a little more
+-- chance of being valid input that has a chance to execute successfully.
+
+-- Right now safelist interaction is not tested, since it seems that doing it
+-- in a useful way require modeling a safelist contract as well.
+
+stablecoinSMTSpec
```
gromakovsky, line 65: Apparently here we test both FA2 and non-FA2 entrypoints (and behavior). But I intentionally created #42 and #43 as two separate issues because I want these tests to be separated. The primary reason is that one day we will move generic FA2 tests outside of this repo to use them in multiple projects.

haskell/test/SMT.hs
```diff
@@ -0,0 +1,808 @@
...
+          Just seedStr -> do
+            let seed = QCGen (Text.Read.read seedStr::SMGen)
+            let args = stdArgs{ replay=Just(seed, 0) }
+            result <- quickCheckWithResult args $ withMaxSuccess qcMaxSuccesses $ smtProperty contracts
+            printSeed result
+          Nothing -> do
+            result <- quickCheckResult $ withMaxSuccess qcMaxSuccesses $ smtProperty contracts
+            printSeed result
+    where
+      printSeed result =
```
gromakovsky, line 82: I figured out that the default seed printing doesn't work because you are using `tasty-hspec` and apparently it loses seed printing somehow. I suggest that we just don't use `hspec` at all as it only causes extra indirection and extra problems. So please remove these hacks used to print seed.

sras: Removed.

ligo/stablecoin/operator.ligo
sras: As of now, after rebasing, this PR does not contain any contracts changes

sras:
@gromakovsky Have rebased and squash old commits.

haskell/test/SMT.hs
sras: Yes, but it seemed to me that unlike Integrational tests, it would be less effective to have FA2 tests isolated here. Because we are running a sequence of contract inputs, and we need both FA2 calls along with the management calls in each of these sequences so that it will bring out bugs that only show up in some combination of these two.

I have separated FA2 tests into a separate commit now. Please check and let me know if we should further isolate it in some manner.

gromakovsky: See how it's done in [morley-ledgers-test](https://gitlab.com/morley-framework/morley/-/tree/master/code/morley-ledgers-test/src/Lorentz/Contracts/Test/SMT). We have SMT for `ApprovableLedger` and for `ManagedLedger`. The former tests only reduced functionality of FA1.2 tokens, but is still useful because it can detect bugs that can happen as result of calling only FA1.2 entrypoints. The latter can find more bugs, but can't be used for any FA1.2 contract.

haskell/test/SMT.hs
```diff
@@ -0,0 +1,666 @@
...
+import Michelson.TypeCheck
+import qualified Michelson.Typed as T
+import qualified Michelson.Untyped as U
+import Util.Named
+
+-- Some implementation notes:
+--
+-- In this module we test the contract by implementing whole contract
+-- functionality in a Haskell, and comparing its behavior against the actual
+-- contract when run using a Michelson interpretr.
```
gromakovsky, line 41:
```suggestion
-- contract when run using a Michelson interpreter.
```

haskell/test/SMT.hs
```diff
@@ -0,0 +1,666 @@
...
+  , gsMasterMinterPool :: [Address]
+  , gsMinterPool :: Map Address Natural
+  , gsPauserPool :: [Address]
+  , gsTokenOwnerPool :: [Address]
+  }
+
+type GeneratorM = ReaderT GeneratorState Gen
+
+instance MonadFail Gen where
+  fail = error "Not implemented"
```
gromakovsky, line 261: It looks fairly hacky to me, can it be avoided? For example, instead of `(owner:_) <- shuffle gsOwnerPool` can you do `owner <- Data.List.head <$> shuffle gsOwnerPool`? Or just use `elements`.

haskell/test/SMT.hs
gromakovsky: Well, actually, since FA2 is still quite unstable, I think current coupling is ok. One day FA2 will become stable and maybe we will implement another FA2 contract. Then we will decouple these tests. But for this PR I think decoupling is not required.

sras:
>I wonder why you implemented state machine testing engine yourself without reusing existing code in morley-ledgers-test. It's not required to use that code, but still sounds like a good idea to me.

I am not sure, but I checked if it makes sense to use it. but I felt that only a small amount of functionality could be inherited. Also, I wanted to try slightly different method. Where as the SMT tests from the library seem to discard the test as soon as there is an error, here I wanted to keep running the inputs, even after the an error,  both on the Michelson model and the haskell model, and compare the accumulated list of errors in both models, and only errorring out in case of the either the contract state or the error list diverging.

gromakovsky:
Okay, no problem. Hopefully later we will have a uniform approach to this kind of tests that will combine benefits of all existing approaches. I've added link to this PR to the [main SMT issue](https://gitlab.com/morley-framework/morley/-/issues/250).

haskell/test/SMT.hs
sras: Have done this using `elements`.

PR #58, opened 2020-05-28
[#56] Permit non existent src for zero transfers

### LINKED ISSUE #56: Permit debit of 0 amount from empty addresses and fix the error message

FA2 clearly states that:
>  If the holder does not hold any tokens of type token_id, the holder's balance is interpreted as zero.

However, our `debit_from` rejects all addresses not present in ledger. It has two bad consequence:
1. Transfer of 0 amount is prohibited from an address with 0 tokens.
2. Transfer of non-zero amount fails with `NOT_IN_LEDGER` instead of `INSUFFICIENT_BALANCE`.

# Acceptance criteria

`debit_from` treats unknown addresses as having 0 balance. Fails with `INSUFFICIENT_BALANCE` for non-zero debit, does nothing otherwise.

sras:
## Description

Fix contract so that it does not fail if the source of a zero amount transfer does not exist. And if the transfer is non-zero, fail with INSUFFICIENT_BALANCE error.

## Related issue(s)

Resolves #56 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/haskell/test/Lorentz/Contracts/Test/FA2.hs
+++ b/haskell/test/Lorentz/Contracts/Test/FA2.hs
@@ -76,9 +76,21 @@ fa2Spec fa2Originate = do
           validate . Right $
             lExpectViewConsumerStorage consumer [balanceExpected]
 
+    it "allows zero transfer from non-existent accounts" $ integrationalTestExpectation $ do
+        let originationParams = addOperator (wallet1, commonOperator) $ defaultOriginationParams
+                { opPermissionsDescriptor = permissionDescriptorOperatorTransferAllowed }
+        withOriginated fa2Originate originationParams $ \fa2contract -> do
+          let
+            transfers = constructSingleTransfer
+              (#from_ .! wallet1)
+              (#to_ .! wallet2)
+              (#amount .! 0)
+
+          withSender commonOperator $ lCallEP fa2contract (Call @"Transfer") transfers
+
+          validate (Right expectAnySuccess)
+
     it "aborts if there is a failure (due to low balance)" $ integrationalTestExpectation $ do
-      -- Tests transactions are applied in order
-      -- Update balances exactly
         let originationParams = addAccount (wallet1, (commonOperators, 10))
               $ addAccount (wallet2, (commonOperators, 0))
               $ addAccount (wallet3, (commonOperators, 0))
@@ -89,7 +101,7 @@ fa2Spec fa2Originate = do
             transfers = constructTransfersFromSender (#from_ .! wallet1)
               [ (#to_ .! wallet2, #amount .! 10)
               , (#to_ .! wallet3, #amount .! 10)
-              , (#to_ .! wallet3, #amount .! 10) -- should fail
+              , (#to_ .! wallet3, #amount .! 10)
               , (#to_ .! wallet5, #amount .! 10)
               ]
 
@@ -98,8 +110,6 @@ fa2Spec fa2Originate = do
           validate $ Left (lExpectAnyMichelsonFailed fa2contract)
 
     it "aborts if there is a failure (due to non existent source account)" $ integrationalTestExpectation $ do
-      -- Tests transactions are applied in order
-      -- Update balances exactly
         let
           originationParams =
               addAccount (wallet1, (commonOperators, 10))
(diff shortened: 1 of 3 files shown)


haskell/test/Lorentz/Contracts/Test/FA2.hs
```diff
@@ -76,6 +76,21 @@ fa2Spec fa2Originate = do
...
             lExpectViewConsumerStorage consumer [balanceExpected]
 
+    it "allows zero transfer from non-existent accounts" $ integrationalTestExpectation $ do
+      -- Tests transactions are applied in order
+      -- Update balances exactly
+        let originationParams = addOperator (wallet1, commonOperator) $ defaultOriginationParams
+                { opPermissionsDescriptor = permissionDescriptorOperatorTransferAllowed }
+        withOriginated fa2Originate originationParams $ \fa2contract -> do
+          let
+            transfers = constructTransfersFromSender (#from_ .! wallet1)
```
awkure, line 86: In this case it's better to use `constructSingleTransfer`

haskell/test/Lorentz/Contracts/Test/FA2.hs
```diff
@@ -76,6 +76,21 @@ fa2Spec fa2Originate = do
...
+              [ (#to_ .! wallet2, #amount .! 0)
+              ]
+
+          withSender commonOperator $ lCallEP fa2contract (Call @"Transfer") transfers
+
+          validate (Right expectAnySuccess)
```
awkure, line 92: Updated ledger needs to be checked

haskell/test/Lorentz/Contracts/Test/FA2.hs
```diff
@@ -292,6 +328,22 @@ fa2Spec fa2Originate = do
...
           validate $ Left (lExpectAnyMichelsonFailed fa2contract)
 
+    it "allows zero transfer from non-existent accounts" $ integrationalTestExpectation $ do
+      -- Tests transactions are applied in order
+      -- Update balances exactly
+        let originationParams = defaultOriginationParams
+                { opPermissionsDescriptor = permissionDescriptorSelfTransferAllowed }
+        withOriginated fa2Originate originationParams $ \fa2contract -> do
+          let
+            transfers = constructTransfersFromSender (#from_ .! wallet1)
```
awkure, line 338: `constructSingleTransfer` here as well as ledger validation need to be applied

haskell/test/Lorentz/Contracts/Test/FA2.hs
sras: @awkure But it does not transfer anything, so what is there to check?

haskell/test/Lorentz/Contracts/Test/FA2.hs
sras: Changed

ligo/stablecoin/actions.ligo
```diff
@@ -141,7 +141,9 @@ function debit_from
       { const updated_balance : nat = abs (ledger_balance - parameter.amount)
       ; updated_ledger := Big_map.update(parameter.from_, Some (updated_balance), updated_ledger)
       }
-  | None -> failwith ("NOT_IN_LEDGER")
+  | None -> if parameter.amount > 0n
+      then failwith ("INSUFFICIENT_BALANCE")
```
gromakovsky, line 145: Isn't it more efficient to assign `current_balance = case store.ledger[parameter.from_] of Some balance -> balance; None -> 0` and then apply debit logic regardless of the amount? I suppose in this case there will be less instruction or maybe about the same amount of instructions, but a relatively long string (`INSUFFICIENT_BALANCE`) will appear only once. Also it will more clearly correspond to the following FA2 statement:
> If the holder does not hold any tokens of type token_id, the holder's balance is interpreted as zero.

sras: Yes, right. Changed.

gromakovsky:
LGTM now. Please rebase to resolve conflicts and squash fixups (I've checked them). Then I will recheck just in case and approve.

sras:
@gromakovsky Have rebased and cleaned up commits..

PR #64, opened 2020-05-29
[#60] Optimize `mint` entrypoint

### LINKED ISSUE #60: Validate sender once in `mint`

Apparently `mint_tokens` validates sender in a loop.
```
  ( (nil : list (operation))
  , List.fold
      ( function
          ( const accumulator : storage
          ; const mint_param  : mint_param
          ) : storage is case accumulator.minting_allowances[senderAddress] of
            None -> (failwith ("NOT_MINTER") : storage)
          | Some (current_allowance) ->
              mint_tokens (accumulator, mint_param, current_allowance)
          end
      , parameters
      , store
      )
  )
```

But it's a bit inefficient because sender stays constant. I believe we can do it more efficiently by extracting sender's minting allowance once and updating only allowance in the loop, not the whole storage.
Also we do `authorize_minter` at the beginning (after `ensure_not_paused`). I suppose we can return `nat` from `authorize_minter` and not check sender anymore.

# Acceptance criteria

`mint` validates sender only once if that's more efficient (I believe it is, but we should check).

awkure:
Problem: We have some unnecessary checks for sender validation that
only increase gas costs and thought the very parameter needs to be
refactored.

Solution: Refactor `mint` entrypoint.

## Related issue(s)

Resolves #60 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/ligo/stablecoin/actions.ligo
+++ b/ligo/stablecoin/actions.ligo
@@ -69,13 +69,14 @@ function authorize_master_minter
     )
 
 (*
- * Authorizes minter and fails otherwise.
+ * Authorizes minter and returns it's current minting allowance.
+ * Fails if sender is not minter.
  *)
 function authorize_minter
   ( const store : storage
-  ) : unit is case store.minting_allowances[Tezos.sender] of
-    Some (u) -> unit
-  | None -> failwith ("NOT_MINTER")
+  ) : nat is case store.minting_allowances[Tezos.sender] of
+    Some (currentAllowance) -> currentAllowance
+  | None -> (failwith ("NOT_MINTER") : nat)
   end
 
 (*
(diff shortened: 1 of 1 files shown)


gromakovsky:
So did you check how it affects contract size and gas costs?

awkure:
Actually contract size is bigger 🤔

gromakovsky:
Okay, then can you revert to the previous version, but still return current minting allowance from `authorize_minter` so that we don't do `accumulator.minting_allowances[senderAddress]` twice? Hopefully at least this change will be beneficial.

gromakovsky:
I was told that even with this new attempt with a smaller and simpler "optimization" the contract size still gets bigger than before the "optimization". I am not sure why, but I suppose we shouldn't spend more time on that, looks like we won't save much gas even if we try hard.

PR #66, opened 2020-06-01
Sras/#63 test for errors

### LINKED ISSUE #63: Check errors in tests

Currently in error scenarios in tests we always use `lExpectAnyMichelsonFailed` and don't expect any specific error. However, latest FA2 version requires some specific errors in some cases. And our specification also specifies some errors for some scenarios. So we should check that correct errors are thrown.

# Acceptance criteria

When we expect some contract action in tests to fail, we should check that the error is the one we expect, not just some arbitrary error.

sras:
## Description

This PR adds checks for expected errors when the contract fails for some reason. Main changes are

1. Add constants to represent the errors.

2. Renamed permissionDescriptorOperatorNoTransfer to permissionDescriptorNoTransfer
This was done the function had value `NoTransfer (#no_transfer .! ())`, which actually denied all transfers, even for the token owner. Apart from that `permissionDescriptorOwnerTransfer` was already defined to represent an configuration that banned operators from doing transfers.

3. Moved configure operator access tests to management tests. This was originally part of FA2 tests. But FA2 does not prescribe to to access check `configure operator` entrypoints. So this checks were moved to management tests, so that it can use the errors defined in the same module.

Apart from this, I was not sure how to test the hook failure errors that are supposed to be thrown from the owner hook implementations. Simply because that happens from the code that is external to this contract.

## Related issue(s)

Resolves #63

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.


haskell/test/Lorentz/Contracts/Test/FA2.hs
```diff
@@ -137,45 +139,49 @@ fa2Spec fa2Originate = do
...
               $ addAccount (wallet4, (commonOperators, 0)) defaultOriginationParams
                 { opPermissionsDescriptor = permissionDescriptorOwnerOrOperatorTransfer }
         withOriginated fa2Originate originationParams $ \fa2contract -> do
           let
-            transfers = constructTransfersFromSender (#from_ .! wallet1)
-              [ (#to_ .! wallet2, #amount .! 10)
-              , (#to_ .! wallet3, #amount .! 10)
-              , (#to_ .! wallet4, #amount .! 10)
-              , (#to_ .! wallet5, #amount .! 10)
+            transfers1 = constructTransfersFromSender (#from_ .! wallet1)
```
awkure, line 152:
```suggestion
            transfers1 = constructSingleTransfer (#from_ .! wallet1) ...
```

haskell/test/Lorentz/Contracts/Test/FA2.hs
```diff
@@ -137,45 +139,49 @@ fa2Spec fa2Originate = do
...
               $ addAccount (wallet4, (commonOperators, 0)) defaultOriginationParams
                 { opPermissionsDescriptor = permissionDescriptorOwnerOrOperatorTransfer }
         withOriginated fa2Originate originationParams $ \fa2contract -> do
           let
-            transfers = constructTransfersFromSender (#from_ .! wallet1)
-              [ (#to_ .! wallet2, #amount .! 10)
-              , (#to_ .! wallet3, #amount .! 0)
-              , (#to_ .! wallet4, #amount .! 10)
-              , (#to_ .! wallet5, #amount .! 10)
+            transfers1 = constructTransfersFromSender (#from_ .! wallet1)
```
awkure, line 174:
```suggestion
            transfers1 = constructSingleTransfer (#from_ .! wallet1) ...
```

haskell/test/Lorentz/Contracts/Test/FA2.hs
```diff
@@ -137,45 +139,49 @@ fa2Spec fa2Originate = do
...
+              [ (#to_ .! wallet2, #amount .! 1)
               ]
+            transfers2 = constructTransfersFromSender (#from_ .! wallet3)
```
awkure, line 177:
```suggestion
            transfers2 = constructSingleTransfer (#from_ .! wallet3) ...
```

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -610,6 +665,12 @@ managementSpec originate = do
             StorageSafelistContract Nothing -> Right ()
             _ -> Left $ CustomValidationError "Token produced malformed storage, this should not have happened"
 
+      it "validates safelist interface" $ integrationalTestExpectation $ do
```
awkure, line 668: I think we could come up with a better description of this test that at least explains that it should fail due to bad safelist parameter

haskell/test/Lorentz/Contracts/Test/FA2.hs
```diff
@@ -904,4 +911,31 @@ fa2Spec fa2Originate = do
...
+fa2ReceiverHookFailed = lExpectFailWith (== [mt|RECEIVER_HOOK_FAILED|])
+
+fa2SenderHookFailed :: ExecutorError -> Bool
+fa2SenderHookFailed = lExpectFailWith (== [mt|SENDER_HOOK_FAILED|])
+
+fa2ReceiverHookUndefined :: ExecutorError -> Bool
+fa2ReceiverHookUndefined = lExpectFailWith (== [mt|RECEIVER_HOOK_UNDEFINED|])
+
+fa2SenderHookUndefined :: ExecutorError -> Bool
+fa2SenderHookUndefined = lExpectFailWith (== [mt|SENDER_HOOK_UNDEFINED|])
```
gromakovsky, line 884: This is fine for now, but when we move these tests to `morley-ledgers-test` (and treat them as generic FA2 tests) we should update this check because according to FA2 an error can be a string or a pair where the first item is a string.

haskell/test/Lorentz/Contracts/Test/FA2.hs
sras: Changed

haskell/test/Lorentz/Contracts/Test/FA2.hs
sras: Changed.

haskell/test/Lorentz/Contracts/Test/FA2.hs
sras: Changed.

haskell/test/Lorentz/Contracts/Test/Management.hs
sras: Have expanded it to say "should fail if parameter of safelist contract does not have the required entrypoints"

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -734,3 +795,56 @@ managementSpec originate = do
...
+
+      withSender commonOperator $ do
+        let operatorParam =
+              (#owner .! wallet1, #operator .! commonOperator)
+
+        let removeOperatorParam = Remove_operator operatorParam
+        lCallEP stablecoinContract (Call @"Update_operators") [removeOperatorParam]
+        validate $ Left mgmNotTokenOwner
+
+
```
gromakovsky, line 850: Looks like too many empty lines btw.

sras: Fixed

gromakovsky:
I believe it can be merged after rebasing and writing more descriptive commit messages.

PR #67, opened 2020-06-02
[#45] Add nettest scenario that covers all entryoints

### LINKED ISSUE #45: Extend nettest scenario to cover all entrypoints

We have a very simple nettest scenario that only calls `transfer`. That's not sufficient, we should at least call all entrypoints to know they don't exceed gas consumption.

# Acceptance criteria

Our nettest-based tests call all entrypoints present in the contract.

Please don't start before #37, #35, #31.

awkure:
Problem: We need to test our stablecoin contract against
real network to ensure that all the entryopints are called
properly.

Solution: Add nettest scenario that calls all entryopints.

## Related issue(s)

Resolves #45 

- Tests
        silently reappearing again.

- Documentation
    - Haddock
  - [X] I updated [changelog](../tree/master/ChangeLog.md) unless I am sure my changes are
        not essential.

### DIFF
--- a/haskell/nettest/Nettest.hs
+++ b/haskell/nettest/Nettest.hs
@@ -5,6 +5,9 @@ module Nettest
   ( scTransferScenario
   ) where
 
+import qualified Data.Set as Set
+
+import qualified Indigo.Contracts.Safelist as SF
 import Lorentz hiding (comment, (>>))
 import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
 import Lorentz.Contracts.Test.Common
@@ -16,31 +19,70 @@ import Util.Named
 import Lorentz.Contracts.Stablecoin
 
 scTransferScenario
-  :: (OriginationParams -> Storage)
+  :: forall m capsM.
+     (Monad m, capsM ~ NettestT m)
+  => (OriginationParams -> Storage)
   -> U.Contract
-  -> NettestScenario
-scTransferScenario constructInitialStorage contract_ = uncapsNettest $ do
+  -> U.Contract -- Note that we do not distinguish safelist and stablecoin contracts
+  -> NettestImpl m
+  -> m ()
+scTransferScenario constructInitialStorage stablecoinContract safelistContract = uncapsNettest $ do
+  comment "Resolving contract managers"
+
   superuser <- resolveNettestAddress
+
+  nettestOwner <- newAddress "nettestOwner"
+  nettestPauser <- newAddress "nettestPauser"
+  nettestMasterMinter <- newAddress "nettestMasterMinter"
+
+  comment "Resolving owners and operators"
 
   owner1 <- newAddress "Steve"
   owner2 <- newAddress "Megan"
   owner3 <- newAddress "Karen"
 
   operator <- newAddress "Some operator"
+  otherOperator <- newAddress "Other operator"
 
   let
     initialStorage =
-        addAccount (superuser , ([operator], 111))
+        addAccount (superuser , ([operator], 1110000))
       . addAccount (owner1, ([operator], 100))
       . addAccount (owner2, ([operator], 0))
       . addAccount (owner3, ([operator], 0))
+      . addAccount (nettestPauser, ([operator], 0))
+      . addMinter (superuser, 200)
       $ defaultOriginationParams
+          { opOwner = nettestOwner
+          , opPauser = nettestPauser
+          , opMasterMinter = nettestMasterMinter
+          }
+
+  comment "Originating stablecoin contract"
 
   scAddress <- do
     let str = constructInitialStorage initialStorage
-    originateUntypedSimple "nettest.Stablecoin" (untypeValue $ toVal str) contract_
+    originateUntypedSimple "nettest.Stablecoin" (untypeValue $ toVal str) stablecoinContract
+
+  comment "Originating safelist contract"
 
   let
+    -- Dummy storage for safelist used in `safelistScenario`
+    initialSafelistStorage = SF.Storage
+      { SF.sTransfers = Set.fromList
+          [ (owner1, owner2)
+          , (owner2, owner3)
+          , (owner3, owner1)
+          ]
+      , SF.sReceivers = Set.fromList [owner1, owner2]
+      }
+
+  sfAddress <- do
+    originateUntypedSimple "nettest.Safelist" (untypeValue $ toVal initialSafelistStorage) safelistContract
+
+  let
+    expectFailed = flip expectFailure NettestFailedWith
+
     sc :: AddressOrAlias
     sc = AddressResolved scAddress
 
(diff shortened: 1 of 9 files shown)


gromakovsky:
Tests are failing because `newAddress` does not send enough XTZ to cover transaction costs. The primary reason is that we use too high transaction costs. It's fixed [here](https://gitlab.com/morley-framework/morley/-/merge_requests/443). I suppose we can wait for that MR to be merged, hopefully it will resolve the problem. If it doesn't (if we make too many calls from some address) you can also manually transfer more from `nettest` to addresses that need more.

haskell/nettest/Nettest.hs
```diff
@@ -54,6 +95,160 @@ scTransferScenario constructInitialStorage contract_ = uncapsNettest $ do
...
+
+    setSafelist :: Address -> Address -> capsM ()
+    setSafelist from safelistAddress = callFrom (AddressResolved from) sc (ep "set_safelist") (Just safelistAddress)
+
+    unsetSafelist :: Address -> capsM ()
+    unsetSafelist from = callFrom (AddressResolved from) sc (ep "set_safelist") (Nothing :: Maybe Address)
+
+  let
+    transferScenario = do
+      comment "Transfer from superuser to superuser "
```
gromakovsky, line 147: IMO it looks a bit confusing when you print `"Transfer from superuser to superuser"`, make a transfer from superuser to superuser and then make other transfers before printing something else. If `callTransfer superuser owner1 40` fails, one will think that "Transfer from superuser to superuser failed", but in reality a different transfer failed. So IMO if there is a comment `X` then all actions between `X` and the next comment `Y` (if there is `Y`) should somehow relate to `X`.

haskell/nettest/Nettest.hs
```diff
@@ -54,6 +95,160 @@ scTransferScenario constructInitialStorage contract_ = uncapsNettest $ do
...
+      callTransfer superuser owner1 40
+
+      callTransfers
+        [ (#from_ .! owner1, [(#to_ .! owner2, #amount .! 10)])
+        , (#from_ .! owner2, [(#to_ .! owner3, #amount .! 10)])
+        , (#from_ .! owner3, [(#to_ .! owner1, #amount .! 10)])
+        ]
+
+      comment "Pausing contract for transfers"
+      expectFailed $ pause owner1 -- Not enough permissions
```
gromakovsky, line 158: One day we'll be able to specify the exact expected error :coolstory:

haskell/src/Indigo/Contracts/Safelist.hs
```diff
@@ -16,6 +16,7 @@ import qualified Data.Set as Set
 
 import Indigo as Indigo
 import Lorentz.Run (Contract)
+import Prelude ()
```
gromakovsky, line 19: Erm, did it become necessary?

haskell/nettest/Nettest.hs
gromakovsky: Can you change it to "Transfers from superuser"? Because there are 2 transfers in this group, both are from `superuser`, but only one is to `superuser`.

awkure:
Note, that I've used a minor [hotfix](https://gitlab.com/morley-framework/morley/-/commit/ec0a1f2d78108c1e82d837b6e1cbd86cf41b5d7e) in morley to make updated integrational tests pass. For the discussion you can refer to our [slack thread](https://serokell.slack.com/archives/CUMAQ3PL4/p1592934064227100)

haskell/test-common/Lorentz/Contracts/Test/Common.hs
```diff
@@ -225,7 +225,7 @@ skipTest = do
     dummyContract = L.unpair L.# L.drop L.# L.nil L.# L.pair
   c <- lOriginate (L.defaultContract dummyContract) "skip test dummy" () (unsafeMkMutez 0)
   lCallEP c CallDefault ()
-  validate (Right expectAnySuccess)
+  expectNoUpdates
```
gromakovsky, line 228: I think `skipTest` is just not needed anymore. It should be possible to just do nothing now, after recent `Michelson.Test` changes.

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -756,8 +759,7 @@ managementSpec originate = do
               ]
 
           withSender wallet1 $ lCallEP stablecoinContract (Call @"Mint") mintings
-
-          validate (Right expectAnySuccess)
+          pass
```
gromakovsky, line 762: All these `pass` usages are redundant, aren't they?

haskell/src/Indigo/Contracts/Safelist.hs
gromakovsky: @awkure ping.

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -471,8 +469,9 @@ managementSpec originate = do
           $ addAccount (wallet1, (commonOperators, 10))
           $ defaultOriginationParams
       withOriginated originate originationParams $ \stablecoinContract -> do
-        withSender wallet1 $ lCallEP stablecoinContract (Call @"Burn") [ 10, 10 ]
-        validate . Left $ mgmInsufficientBalance
+        -- withSender wallet1 $ lCallEP stablecoinContract (Call @"Burn") [ 10, 10 ]
```
gromakovsky, line 472: Looks like a leftover.

haskell/src/Indigo/Contracts/Safelist.hs
awkure: Yes, because it conflicts with `import Indigo` recently updated operators

gromakovsky: Alright, maybe it would be better to enable `NoImplicitPrelude` globally for the lib, but I am not sure, so let's not do it for now.

haskell/test/Lorentz/Contracts/Test/FA2.hs
```diff
@@ -607,9 +586,57 @@ fa2Spec fa2Originate = do
             let isOperatorQuery = mkView (#operator .! operatorParam) consumer
             lCallEP fa2contract (Call @"Is_operator") isOperatorQuery
 
-            (validate . Right $
-              lExpectViewConsumerStorage consumer
-                [(#operator .! operatorParam, #is_operator .! True)])
+            lExpectViewConsumerStorage consumer
+                [(#operator .! operatorParam, #is_operator .! True)]
+
+  -- Check whether "update operator", "remove operator" operations are executed only by contract owner.
```
gromakovsky, line 592: Are these tests new? Why did you add them as part of morley update?

haskell/nettest/Nettest.hs
```diff
@@ -50,10 +92,186 @@ scTransferScenario constructInitialStorage contract_ = uncapsNettest $ do
...
+    callTransfers = mapM_ $ \(from@(arg #from_ -> from_), destinations) ->
+      callFrom
+        (AddressResolved from_)
+        sc
+        (ep "transfer")
+        (constructTransfersFromSender from destinations)
+
+    configureMinter :: Address -> Address -> Maybe Natural -> Natural -> capsM ()
+    configureMinter from for expectedAllowance newAllowance =
+      callFrom (AddressResolved from) sc (ep "configure_minter") (#minter .! for, (#current_minting_allowance .! expectedAllowance, #new_minting_allowance .! newAllowance))
```
gromakovsky, line 112: Some lines here are :so-big:, this one is 173 characters long, it's not ok.

awkure: :bruh:

gromakovsky:
Also please check whether there are warnings. IIRC CI doesn't check that for some reason.

haskell/test/Lorentz/Contracts/Test/FA2.hs
awkure: Well, this is weir how this appeared in management tests and got merged onto master. I'll make a separate commit for it

gromakovsky: And where is this promised separate commit? :pled:

Issue #52, opened 2020-05-27
Add weeder and hlint checks to CI


gromakovsky:
# Clarification and motivation

We normally want our Haskell code to pass weeder and hlint checks, but this repo doesn't check it on CI.

# Acceptance criteria

CI should fail if Haskell code doesn't pass `hlint` (with our configs) and `weeder` checks.

Note that there is a new [weeder](https://github.com/ocharles/weeder), maybe it's better to use it instead of the old one. The old one is more powerful (it detects unused deps), but it's not deprecated AFAIU. We can even run both, so that we detect as much as possible now and will remove the old one when it's totally dead.

gromakovsky:
Also please cleanup CI config, it has useless lines:
```
# TODO: rewrite it to haskell.nix

env:
  NIX_PATH: nixpkgs=https://github.com/nixos/nixpkgs/archive/b640dbd00877f51616750e347a32fbcb49d2968c.tar.gz
```

rvem:
We'll add new weeder to the CI pipeline in a separate issue once the old weeder will become deprecated

PR #70, opened 2020-06-05
[#52] update CI pipeline

rvem:
## Description
Add hlint and weeder steps to the CI pipeline.

TODO: figure out whether we want to use new weeder as well.

## Related issue(s)

Resolves partially #52 

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -31,5 +31,5 @@ in
   lib = project.stablecoin.components.library;
   test = project.stablecoin.components.tests.stablecoin-test;
   nettest = project.stablecoin.components.tests.stablecoin-nettest;
-  inherit tezos-contract tezos-client;
+  inherit tezos-contract tezos-client pkgs;
 }
--- a/default.nix
+++ b/default.nix
@@ -4,16 +4,26 @@
 { sources ? import ./nix/sources.nix
 , haskell-nix ? import sources."haskell.nix" { }
 , pkgs ? import sources.nixpkgs haskell-nix.nixpkgsArgs
+, weeder-hacks ? import sources.haskell-nix-weeder { inherit pkgs; }
 , ligo ? (import "${sources.ligo}/nix" { }).ligo-bin
 }:
 let
+  local-packages = [{
+      name = "stablecoin";
+      subdirectory = ".";
+  }];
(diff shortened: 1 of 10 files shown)


default.nix
```diff
@@ -4,16 +4,29 @@
...
+        packages = pkgs.lib.genAttrs local-packages-names (packageName: {
+            package.ghcOptions = with pkgs.lib;
+              concatStringsSep " " ([
+                "-ddump-to-file" "-ddump-hi"
+              ]);
+            postInstall = weeder-hacks.collect-dump-hi-files;
+        });
+        # packages.stablecoin.components.tests.stablecoin-test = {
+        #   postInstall = "mkdir -p $out/test/resources; cp -Lr ${tezos-contract} $out/test/resources/stablecoin.tz";
+        # };
```
gromakovsky, line 29: Is it needed?

gromakovsky:
> Resolves partially #52

Because of new weeder?

> TODO: figure out whether we want to use new weeder as well.

Did you try doing that? I would try and if it appears to be straightforward – let's do it, otherwise let's forget about it for now. I suppose one day we will have to switch to new `weeder` (or just stop using the old one because it's not maintained, but not start using the new one), so our first step should be to check whether it's easy to setup.

rvem:
> Because of new weeder?
Yes
> Did you try doing that?
Currently, I'm trying to build it and test locally

rvem:
```
$ ./weeder
src/Indigo/Contracts/Safelist.hs:28:1: error: mkStorage is unused

      26 weeder: <stdout>: commitBuffer: invalid argument (invalid character)
```
Sad that it doesn't produce fancy output like in https://github.com/ocharles/weeder#calling-weeder

rvem:
okay, looks like new weeder requires the presence of `.hie` files, thus we will need some additional workarounds to make it work with haskell.nix, let's do it in a separate PR

gromakovsky:
> Resolves partially #52

@rvem why partially? Because of new `weeder`? I think it's not a big problem, using only one `weeder` should be fine. We can get back to this question one the old `weeder` dies.

rvem:
Okay, let's return to this once the old weeder is deprecated

PR #71, opened 2020-06-08
[#48] morley optimizer integration

### LINKED ISSUE #48: Incorporate optimizer into our pipeline

[Morley](https://gitlab.com/morley-framework/morley) has an optimizer that can turn a contract into an equivalent one with smaller size and gas consumption. We should use it to make our contract smaller.

# Acceptance criteria

1. `Makefile` exposes some way to build stablecoin contract and apply `morley optimize` command to it.
2. Our tests use the optimized version of the contract.

rvem:
## Description

This PR adds an ability to optimize generated ligo contract via `morley-optimize`.
An optimized contract is used as a test resource in CI.

## Related issue(s)

Resolves #48 

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,6 +1,10 @@
 # SPDX-FileCopyrightText: 2020 tqtezos
 # SPDX-License-Identifier: MIT
 
+env:
+  FETCH_CONTRACT:
+    "mkdir -p test/resources/ && buildkite-agent artifact download stablecoin.tz test/resources/ --step \"LIGO-contract\""
+
 steps:
  - label: hlint
    if: &not_scheduled
@@ -20,11 +24,17 @@ steps:
  - label: build library
    if: *not_scheduled
    command: nix build -L -f. lib
+ - label: LIGO-contract
+   commands:
+     - nix-build -A tezos-contract -o stablecoin_raw.tz
+     - nix run -f. morley -c morley optimize --contract stablecoin_raw.tz --output stablecoin.tz
+   artifact_paths:
+     - stablecoin.tz
  - label: test
    if: *not_scheduled
    commands:
+     - eval "$FETCH_CONTRACT"
      - nix build -L -f. test
-     - nix-build -A tezos-contract -o test/resources/stablecoin.tz
      - ./result/bin/stablecoin-test
  - label: nettest-local-chain
    if: *not_scheduled
@@ -35,8 +45,7 @@ steps:
      MONEYBAG: "unencrypted:edsk3GjD83F7oj2LrnRGYQer99Fj69U2QLyjWGiJ4UoBZNQwS38J4v"
    commands: &run-nettest
      - nix build -L -f. nettest
-     - mkdir -p test/resources/
-     - nix-build -A tezos-contract -o test/resources/stablecoin.tz
+     - eval "$FETCH_CONTRACT"
      - nix run -f. tezos-client -c ./scripts/ci/run-local-chain-nettest.sh result/bin/stablecoin-nettest
    retry: &retry-nettest
       automatic:
@@ -52,9 +61,6 @@ steps:
      MONEYBAG: "unencrypted:edsk4EYgVvKVKppJXAyeoqxJrNeyoktNqmcx5op1CFht9P1p8pPxp7"
    commands: *run-nettest
    retry: *retry-nettest
- - label: check LIGO contract
-   if: *not_scheduled
-   command: nix build -L -f. tezos-contract
  - label: weeder
    if: *not_scheduled
    commands:
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,11 @@
 # SPDX-FileCopyrightText: 2020 tqtezos
 # SPDX-License-Identifier: MIT
 
-.PHONY: all build-ligo build-haskell test test-dumb-term test-hide-successes nettest clean
+.PHONY: all build-ligo build-haskell optimize-ligo test test-dumb-term test-hide-successes nettest clean
+
+# Morley executable, it is used for optimizing Michelsion version
+# of stablecoin.tz
+MORLEY ?= morley
 
 MAKE_HASKELL = $(MAKE) -C haskell/
 MAKE_LIGO = $(MAKE) -C ligo/
@@ -15,6 +19,10 @@ build-ligo:
 # Build everything haskell-related (including tests and benchmarks) with development options.
 build-haskell:
 	$(MAKE_HASKELL) build
+
+# Optimize built ligo contract using morley optimizer
+optimize-ligo:
+	$(MORLEY) optimize --contract ligo/stablecoin.tz --output ligo/stablecoin.tz
 
 test:
 	$(MAKE_HASKELL) test
--- a/default.nix
+++ b/default.nix
@@ -14,7 +14,10 @@ let
   }];
   local-packages-names = map (p: p.name) local-packages;
   project = pkgs.haskell-nix.stackProject {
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -1,7 +1,10 @@
 # SPDX-FileCopyrightText: 2020 tqtezos
 # SPDX-License-Identifier: MIT
 
-.PHONY: all build-ligo build-haskell test test-dumb-term test-hide-successes nettest clean
+.PHONY: all build-ligo build-haskell optimize-ligo test test-dumb-term test-hide-successes nettest clean
+
+# Ligo executable
```
gromakovsky, line 6: :jebaited:

Makefile
```diff
@@ -16,6 +19,10 @@ build-ligo:
 build-haskell:
 	$(MAKE_HASKELL) build
 
+# Optimitez built ligo contract using morley optimizer
```
gromakovsky, line 22: > Optimitez

Wut?

haskell/Makefile
```diff
@@ -28,7 +29,7 @@ endef
 define build_stablecoin_contract
 	$(MAKE_LIGO) stablecoin.tz
 	mkdir -p test/resources
-	ln -sf "../../$(LIGO_DIR)/stablecoin.tz" "test/resources/stablecoin.tz"
+	morley optimize --contract "$(LIGO_DIR)/stablecoin.tz" --output "test/resources/stablecoin.tz"
```
gromakovsky, line 32: You define `MORLEY` but use `morley`.

nix/sources.json
```diff
@@ -40,5 +40,16 @@
...
         "url": "https://github.com/serokell/tezos-packaging/archive/8e7e651e5a02858ffbb9900fb200498777651df6.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
+    },
+    "morley": {
+        "branch": "master",
+        "description": "Developer tools for the Michelson Language",
+        "homepage": "",
+        "owner": "morley-framework",
+        "repo": "morley",
+        "rev": "f61e31dfd4977350cbc1f3fb23392c8799bc2fe4",
```
gromakovsky, line 50: Erm, it's very sad that now `morley` revision appears in two places. Strictly speaking, it doesn't have to match the one from `stack.yaml`, so it's not that big of a problem, but still sad.
1. Can we somehow extract this information from `stack.yaml` instead?
2. If we can't, I find it better to use `pkgs.haskellPackages.morley`. Right now `morley` on Hackage has an issue that it depends on `tezos-bake-monitor-lib`, but we are very close to eliminating this dependency. Once we finish eliminating it, we will make a new release which will hopefully be usable via `pkgs.haskellPackages.morley`, at least that's how I understand it.

.buildkite/pipeline.yml
```diff
@@ -35,7 +37,8 @@ steps:
    commands: &run-nettest
      - nix build -L -f. nettest
      - mkdir -p test/resources/
-     - nix-build -A tezos-contract -o test/resources/stablecoin.tz
+     - nix-build -A tezos-contract -o stablecoin.tz
+     - nix run -f. morley -c morley optimize --contract stablecoin.tz --output test/resources/stablecoin.tz
```
gromakovsky: Now I see some non-trivial amount of duplication between this step and the one where we run `stablecoin-test`. Is it possible to deduplicate common parts?

Makefile
rvem: I've tried :joy:

Makefile
rvem: :yao:

nix/sources.json
rvem: > Can we somehow extract this information from stack.yaml instead?

Looks hacky, nix dependencies doesn't have anything to do with stack dependencies

gromakovsky:
>  fixup! [#48] Add optimitez call to the Makefile 

You could have used `--squash` here btw. In that case during `git rebase --autosquash` you would be able to fix the commit message :)

.buildkite/pipeline.yml
```diff
@@ -4,6 +4,8 @@
 
 env:
   NIX_PATH: nixpkgs=https://github.com/nixos/nixpkgs/archive/b640dbd00877f51616750e347a32fbcb49d2968c.tar.gz
+  FETCH_CONTRACT:
+    "mkdir -p test/resources/ && buildkite-agent artifact download stablecoin.tz test/resources/ --step \"LIGO contract\""
```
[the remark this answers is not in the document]
rvem: I think that it's third time, when I'm upset because of list concatenation absence

nix/sources.json
gromakovsky: https://gitlab.com/morley-framework/morley/-/merge_requests/456

gromakovsky: Can you try `morley-1.4.0`?

gromakovsky:
> error: Package ‘morley-0.6.0’ in /nix/store/y38lsywxy4dnpmnlmpgm1fndjcx87h5y-source/pkgs/development/haskell-modules/hackage-packages.nix:167947 is marked as broken, refusing to evaluate.

Why is it trying 0.6.0? That's fairly old.

gromakovsky:
Also I suggest rebasing.

rvem:
> Why is it trying 0.6.0?
That's because of haskell.nix adds haskell packages that are in lts-15.13. However, morley isn't there, so it's using very outdated morley that appeared in nixpkgs.

.buildkite/pipeline.yml
```diff
@@ -35,8 +46,7 @@ steps:
      MONEYBAG: "unencrypted:edsk3GjD83F7oj2LrnRGYQer99Fj69U2QLyjWGiJ4UoBZNQwS38J4v"
    commands: &run-nettest
      - nix build -L -f. nettest
-     - mkdir -p test/resources/
-     - nix-build -A tezos-contract -o test/resources/stablecoin.tz
+     - bash -c "$FETCH_CONTRACT"
```
gromakovsky, line 49: Isn't it better to call `/usr/bin/env bash` here and above? It probably shouldn't affect anything, just a matter of style.

.buildkite/pipeline.yml
```diff
@@ -20,11 +24,18 @@ steps:
  - label: build library
    if: *not_scheduled
    command: nix build -L -f. lib
+ - label: LIGO-contract
+   if: build.source != "schedule"
```
gromakovsky, line 28: Why not `   if: *not_scheduled`?

gromakovsky: Also are we sure it's not needed for scheduled builds?

rvem: hmm, this actually looks like a rebase artifact

.buildkite/pipeline.yml
rvem: I think that `eval "$FETCH_CONTRACT"` better suits here, dunno, why I chose `bash -c` here

default.nix
```diff
@@ -39,11 +42,15 @@ let
     hs-pkgs = project;
     local-packages = local-packages;
   };
+  morley =
+    (pkgs.haskell-nix.hackage-package
+      { name = "morley"; version = "1.4.0"; compiler-nix-name = "ghc883"; }
```
gromakovsky, line 47: Btw, I wonder which optimization settings will be used when `morley` is built.

rvem: AFAIU, cabal-to-nix uses `cabal v2-configure` which has `-O1` by default

PR #73, opened 2020-06-12
[#68] Address to recent FA2 changes (part 2)

### LINKED ISSUE #68: Account for new FA2 updates

There are new MRs updating FA2:
* https://gitlab.com/tzip/tzip/-/merge_requests/66
* https://gitlab.com/tzip/tzip/-/merge_requests/67

Once they are merged, we should update our code accordingly.

# Acceptance criteria

Once the above MRs are both merged our specification, contract code and tests are updated accordingly.

awkure:
## Description

This PR updates stablecoin contract according to recent FA2 changes, specifically [!66](https://gitlab.com/tzip/tzip/-/merge_requests/66) and [!67](https://gitlab.com/tzip/tzip/-/merge_requests/67)

But note, that [!67](https://gitlab.com/tzip/tzip/-/merge_requests/67) is about updating transfer parameter which is done already here and mostly 
for their inner code conversion refactoring and document restructure, so only [!66](https://gitlab.com/tzip/tzip/-/merge_requests/66) is left to be
done.

## Related issue(s)

Resolves #68 

- Tests
        silently reappearing again.

- Documentation
    - Haddock
  - [X] I updated [changelog](../tree/master/ChangeLog.md) unless I am sure my changes are
        not essential.

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -92,11 +92,11 @@ In error scenarios the stablecoin contract fails with a string.
 Here is a summary of all the strings used as error messages.
 We start with standard FA2 errors which are part of the FA2 specification.
 
-| Error                  | Description |
-|------------------------|-------------|
-| `TOKEN_UNDEFINED`      | One of the specified `token_id`s is not defined (i. e. not zero) |
-| `INSUFFICIENT_BALANCE` | Cannot debit from a wallet because of excessive amount of tokens |
-| `NOT_OPERATOR`         | A transfer is initiated neither by the token owner nor a permitted operator |
+| Error                      | Description                                                                 |
+|----------------------------|-----------------------------------------------------------------------------|
+| `FA2_TOKEN_UNDEFINED`      | One of the specified `token_id`s is not defined (i.e. not zero)             |
+| `FA2_INSUFFICIENT_BALANCE` | Cannot debit from a wallet because of excessive amount of tokens            |
+| `FA2_NOT_OPERATOR`         | A transfer is initiated neither by the token owner nor a permitted operator |
 
 The next group consists of the errors that are not part of the FA2 specification.
 
@@ -648,7 +648,7 @@ Parameter (in Michelson): `list nat`.
 
 - Fails with `NOT_MINTER` if the sender is not a minter.
 
-- Fails with `INSUFFICIENT_BALANCE` if the sender does not have enough tokens to burn.
+- Fails with `FA2_INSUFFICIENT_BALANCE` if the sender does not have enough tokens to burn.
 
 - A minter with 0 minting allowance is allowed to burn tokens.
 
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -73,6 +73,12 @@ Each address can have any number of operators and be an operator of any number o
 Every address that is stored in ledger is associated with its current balance.
 Additionally, each minter is associated with its current minting allowance.
 
+## Token metadata
+
+Stablecoin contract MUST provide an additional metadata `big_map` for its tokens.
+This `big_map` MUST be annotated as `%token_metadata` and can be at any position within
+contract storage. The structure of this `big_map` MUST follow [FA2 requirements](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-12/tzip-12.md#token_metadata-big_map).
+
 # Safelist
 
 Before describing the token contract's entrypoints we describe the `Safelist` interface.
@@ -127,8 +133,7 @@ Since they are internal, they can be changed any time without updating this part
 Full list:
 * [`transfer`](#transfer)
 * [`balance_of`](#balance_of)
-* [`total_supply`](#total_supply)
-* [`token_metadata`](#token_metadata)
+* [`token_metadata_registry`](#token_metadata_registry)
 * [`permissions_descriptor`](#permissions_descriptor)
 * [`update_operators`](#update_operators)
 * [`is_operator`](#is_operator)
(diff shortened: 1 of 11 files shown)


gromakovsky:
AFAIU, currently it only accounts for !67, right? I see that !67 was closed and `total_supply` was removed in !66. So most likely it will be removed eventually.

Apart from that note that !66 was is not merged yet, so maybe some changes will happen there, but most likely there will be no substantial changes, so I think you can already start incorporating !66 changes, but we won't merge them before they land to tzip's master.

awkure:
Oh, I mean that we have already done !67 earlier and this MR accommodates changes from !66 (which is actually closed but not merged) so I believe that this issue may be closed as well, but it needs to be re-checked.

awkure:
Ok, `token_metadata` type seems to be updated and an additional entrypoint `Token_metadata_registry` was added. I wish they at least explain what they have done somewhere before merging these changes to master.

gromakovsky:
Actually there are quite a lot of changes in !66. And it was merged recently btw.

docs/specification.md
```diff
@@ -324,7 +325,7 @@ token_metadata_response =
 
 token_metadata =
```
sras, line 326: Here and above `token_metadata` type alias is used. I think one of them should be `token_metadata_request`.

sras: Also, do we need this entrypoint? Because aren't we implementing this alternative? 

>Contract storage MUST have a big_map that maps token_id -> token_metadata
and annotated %token_metadata

haskell/test/Lorentz/Contracts/Test/FA2.hs
```diff
@@ -844,7 +786,7 @@ fa2Spec fa2Originate = do
 
           withSender commonOperator $ lCallEP fa2contract (Call @"Transfer") transfers
 
-          validate $ Left fa2SenderHookUndefined
+          validate $ Left fa2SenderHookFailed
```
sras, line 789: Why was undefined hook error changed to hook failed error?

docs/specification.md
```diff
@@ -129,6 +129,7 @@ Full list:
 * [`balance_of`](#balance_of)
 * [`total_supply`](#total_supply)
```
sras, line 130: Shouldn't we remove `total_supply` from here and the corresponding section below as well?

docs/specification.md
```diff
@@ -324,7 +325,7 @@ token_metadata_response =
 
 token_metadata =
   ( list token_id                           :token_ids
-  , contract (list token_metadata_response) :callback
+  , contract (list token_metadata) :callback
```
sras, line 328: Shouldn't this be a lambda instead?

docs/specification.md
```diff
@@ -766,6 +783,8 @@ Note that operator relation is not transitive.
 
 One more uncertainty is related to the `token_metadata` entrypoint: whether metadata is constant or can be changed during contract's lifetime.
 We assume that token metadata stays constant forever.
+Upd: We have an additional entrypoint that holds all token metadata
```
sras, line 786: >Upd:

Not sure if this is a left over.

haskell/test-common/Lorentz/Contracts/Test/Common.hs
```diff
@@ -211,6 +212,14 @@ constructSingleTransfer
 constructSingleTransfer from to amount
     = [(from, (#txs .! [(to, (#token_id .! 0, amount))]))]
 
+-- | Token metadata parameter that takes list of token ids and
+-- assumes that each of which resulting metadata is always valid
+assertAnyTokenMetaData :: [Natural] -> FA2.TokenMetaDataParam
+assertAnyTokenMetaData ids =
```
sras, line 218: I am not sure if it is proper to call this an assertion, because it can be something else, like posting token metadata to a different contract. right?

docs/specification.md
awkure: No, it's an commentary that  needs to be discussed

haskell/test/Lorentz/Contracts/Test/FA2.hs
awkure: To match `FA2_RECEIVER_HOOK_FAILED` error as is defined in FA2 spec 🤷

docs/specification.md
awkure: Well, If it's required by FA2 so I believe that we should leave it as is here as well

docs/specification.md
awkure: yep

haskell/test/Lorentz/Contracts/Test/FA2.hs
sras: But here the hook is undefined here on the reciever, so shouldn't it be the undefined error as per FA2?

docs/specification.md
sras: But AFAIU FA2 specifies

>Token registry contract MUST implement one of two ways to expose token metadata
for off-chain clients:

So it can be either one of those. I ask because @gromakovsky suggested here [1] that we can use the simpler alternative of just leaving the metadata in the storage, without actually implementing the entrypoint alternative.

[1] https://serokell.slack.com/archives/CUMAQ3PL4/p1592391873179000

awkure: Alright, have removed this entrypoint completely

haskell/test/Lorentz/Contracts/Test/FA2.hs
awkure: Ah, you're right, renamed back

haskell/test/SMT.hs
```diff
@@ -7,29 +7,30 @@ module SMT
   ) where
 
 import qualified Data.Map as Map
+import Data.Maybe (fromJust)
```
gromakovsky, line 10: Is it used here?

ligo/stablecoin/spec.ligo
```diff
@@ -79,8 +79,7 @@ end
 type balance_of_params is michelson_pair_right_comb(balance_of_params_)
 
 type token_metadata_ is record
-  token_id  : token_id
```
gromakovsky, line 96: Why did you remove `token_id` from here? AFAIU it contradicts FA2. We even have discussed it with Sandeep in [this thread](https://serokell.slack.com/archives/CUMAQ3PL4/p1592391873179000) before.

haskell/test/SMT.hs
gromakovsky: @awkure ping

haskell/test/SMT.hs
```diff
@@ -7,29 +7,30 @@ module SMT
...
 import Data.Typeable (cast)
 import Fmt
```
gromakovsky, line 12: This import should be explicitl

gromakovsky:
You can cleanup existing commits btw, I have checked them (now there are some commits that fix previous ones and should be squashed).

docs/specification.md
```diff
@@ -271,90 +270,21 @@ Parameter (in Michelson):
...
-```
-
-- This entrypoint MUST follow the FA2 requirements.
-
-- Since the contract supports only a single token type, all `token_id` values MUST be 0.
-  They are passed because FA2 requires that.
+- Return contract address that holds token metadata.
 
-- Token metadata for 0 `token_id` is constant and should be hardcoded in contract code or put into storage during origination.
+- Since the contract owns its token metadata the returned value of this entrypoint call will always be equal to `SELF`.
```
gromakovsky, line 287: I guess we should write somewhere that this contract (i.e. our stablecoin contract) must have a `big_map` with a certain annotation and type.

gromakovsky:
```
Problem: With recent FA2 changes `token_metadata` was restructured
so that its parameter accepts `handler` lambda rather than
a callback contract. Also `token_metadata_registry` entryopint
was added that returns an address of a contract that stores actual
token metadata.

Solution: Add `token_metadata_registry` entrypoint, restructure
`token_metadata` parameter and update specification and hasskell
types with tests as well.
```

1. ↑ It's a bit outdated now.
2. There are some unresolved discussions above.
3. Minor: the last commit has some changes that fix previous commits:
  + Restoration of some changed errors (amends the first commit)
  + Removal of `total_supply` from the specification.

gromakovsky:
Also it needs to be rebased.

gromakovsky:
Alright, waiting for rebase so that I can take a look at the contract from CI.

PR #76, opened 2020-07-09
[#75] Transferlist

### LINKED ISSUE #75: Incorporate latest transferlist

There is [this smart contract](https://github.com/tqtezos/lorentz-contract-transferlist), it implements [TZIP-15](https://gitlab.com/tzip/tzip/-/blob/50ba074e1361b470f4f50217794f651168c2f58d/proposals/tzip-15/tzip-15.md). We want our `stablecoin` smart contract to work with that smart contract.

# Acceptance criteria

1. In the specification, Safelist is renamed to Transferlist (to match TZIP-15).
2. In our code, Safelist is renamed as well.
3. Our tests for safelist interaction are updated so that they test both implementations: our simple one (written in Indigo) and the one from `lorentz-contract-transferlist`. We can compile `lorentz-contract-transferlist` to Michelson and add `transferlist.tz` to this repo (to be used in tests).

awkure:
## Roadmap

- [X] Renamings 
- [X] Update types 
- [X] Define `lorentz-contract-whitelist` types and helpers
- [X] Integrational
- [x] Nettest 
- [x] Minors

## Related issue(s)

Resolves #75 

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -45,8 +45,9 @@ steps:
      MONEYBAG: "unencrypted:edsk3GjD83F7oj2LrnRGYQer99Fj69U2QLyjWGiJ4UoBZNQwS38J4v"
    commands: &run-nettest
      - nix build -L -f. nettest
+     - cd haskell/
      - eval "$FETCH_CONTRACT"
-     - nix run -f. tezos-client -c ./scripts/ci/run-local-chain-nettest.sh result/bin/stablecoin-nettest
+     - nix run -f ../ tezos-client -c ../scripts/ci/run-local-chain-nettest.sh ../result/bin/stablecoin-nettest
    retry: &retry-nettest
       automatic:
         limit: 1
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,6 @@ TAGS
 
 # We ignore Michelson files in `ligo/` directory because they are auto-generated.
 /ligo/*.tz
+
 # We also ignore them in `haskell/test/resources/` because they are symlinked for tests.
-/haskell/test/resources/*.tz
+/haskell/test/resources/stablecoin.tz
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,9 +4,14 @@ SPDX-License-Identifier: MIT
 -->
 # Changelog for stablecoin
 
+## 1.1.0
+* [#75](Incorporate safelist contract changes according to TZIP-15):
+  1. Safelist renamed to "Transferlist".
+  2. Transferlist interaction updated to be able to use external transferlists.
+
 ## 1.0.0
 
-* [#73] Update the contract to a newer FA2 revision:
+* [#73](Update the contract to a newer FA2 revision):
   1. No `total_supply`.
   2. New way of exposing metadata.
   3. Slightly modified error messages.
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -33,8 +33,8 @@ Note that the actual storage type is implementation detail and is not specified 
 
 * The contract maintains a ledger of addresses and manages some special roles as described below.
 * Token operations can be paused, so the contract also knows whether it is paused.
-* Token operations can be guarded by the [`Safelist` contract](#safelist), so the contract also optionally stores an address of such contract.
-If its value is `None`, no safelist checks are performed.
+* Token operations can be guarded by the [`Transferlist` contract](#transferlist), so the contract also optionally stores an address of such contract.
+If its value is `None`, no transferlist checks are performed.
 
 ## Roles
 
@@ -79,18 +79,18 @@ Stablecoin contract MUST provide an additional metadata `big_map` for its tokens
 This `big_map` MUST be annotated as `%token_metadata` and can be at any position within
 contract storage. The structure of this `big_map` MUST follow [FA2 requirements](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-12/tzip-12.md#token_metadata-big_map).
 
-# Safelist
+# Transferlist
 
-Before describing the token contract's entrypoints we describe the `Safelist` interface.
+Before describing the token contract's entrypoints we describe the `Transferlist` interface.
 The syntax is: `entrypointName(argumentType)`.
-Required `Safelist` entrypoints:
+Required `Transferlist` entrypoints:
 * `assertTransfers(list (pair (address %from) (list %tos address)))`
   * Checks whether a transfer is permitted from one address to the other one.
   Fails if transfer is prohibited for any item from this list.
-* `assertReceiver(address)`
-  * Fails if address is not whitelisted or it is blacklisted.
 * `assertReceivers(list address)`
   * Fails if any address in the list is not whitelisted or is blacklisted.
+
+The specific interface of transferlist contracts is described in [TZIP-15](https://gitlab.com/tzip/tzip/-/blob/c38f4c2dc9da883d01e85409d34d0015d6ad272a/proposals/tzip-15/tzip-15.md)
 
 # Errors
 
@@ -122,7 +122,7 @@ The next group consists of the errors that are not part of the FA2 specification
 | `ALLOWANCE_MISMATCH`         | In `configure_minter` both allowances are not `None`, but different                                                            |
 | `ADDR_NOT_MINTER`            | An attempt is made to modify minter data of an address that's not a minter                                                     |
 | `ALLOWANCE_EXCEEDED`         | Throws when trying to mint tokens more than currently allowed for an address                                                   |
-| `BAD_SAFELIST`               | Given address is a not a smart contract complying with the safelist interface                                                  |
+| `BAD_TRANSFERLIST`           | Given address is a not a smart contract complying with the transferlist interface                                                  |
 
 Finally there are some internal errors that should be considered implementation detail and are supposed to never happen as long as the contract is originated correctly (with consistent storage).
 Since they are internal, they can be changed any time without updating this part of the specification.
@@ -147,7 +147,7 @@ Full list:
 * [`accept_ownership`](#accept_ownership)
 * [`change_master_minter`](#change_master_minter)
 * [`change_pauser`](#change_pauser)
-* [`set_safelist`](#set_safelist)
+* [`set_transferlist`](#set_transferlist)
 
 Format:
 ```
@@ -219,7 +219,7 @@ Parameter (in Michelson):
 
 - Permission logic requirements specific to this contract are described in the ["FA2 Specifics"](#fa2-specifics) chapter.
 
-- If safelist contract is set, this entrypoint MUST call its `assertTransfers` entrypoint (either directly or from the transfer hook).
+- If transferlist contract is set, this entrypoint MUST call its `assertTransfers` entrypoint (either directly or from the transfer hook).
 
 - Since the contract supports only a single token type, all `token_id` values MUST be 0.
   They are passed because FA2 requires that.
@@ -567,7 +567,7 @@ Parameter (in Michelson):
 - Minting allowance will decrease by the amount of tokens minted
   and increase the balance of receiver and total minted value.
 
-- If safelist contract is set, this entrypoint MUST call its `assertReceivers` entrypoint (either directly or from the transfer hook) checking the minter and all receivers.
+- If transferlist contract is set, this entrypoint MUST call its `assertReceivers` entrypoint (either directly or from the transfer hook) checking the minter and all receivers.
 
 - Fails with `CONTRACT_PAUSED` if the contract is paused.
 
@@ -590,7 +590,7 @@ Parameter (in Michelson): `list nat`.
 
 - Burning tokens will not increase the mintingAllowance of the address doing the burning.
 
-- If safelist contract is set, this entrypoint MUST call its `assertReceiver` entrypoint (either directly or from the transfer hook) checking the minter (`SENDER`).
+- If transferlist contract is set, this entrypoint MUST call its `assertReceivers` entrypoint (either directly or from the transfer hook) checking the minter (`SENDER`).
 
 - Fails with `CONTRACT_PAUSED` if the contract is paused.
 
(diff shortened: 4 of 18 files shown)


docs/specification.md
```diff
@@ -92,6 +92,8 @@ Required `Safelist` entrypoints:
 * `assertReceivers(list address)`
   * Fails if any address in the list is not whitelisted or is blacklisted.
 
+We currently support 2 versions of transferlist: [Ours](https://github.com/tqtezos/stablecoin/blob/bd3c59cf3d084a88d1695829d6f2015018dc5cbc/haskell/src/Indigo/Contracts/Transferlist.hs) and Michael's [lorentz-contract-transferlist](https://github.com/tqtezos/lorentz-contract-transferlist/tree/476c0429adf74041ed7d26c44e5e8e73f5a54f36) located at TODO
```
gromakovsky, line 95: We support any version of Transferlist, as long as its has three entrypoints mentioned above. We can mention TZIP-15 here and say that these entrypoints are from TZIP-15.

.gitignore
```diff
@@ -14,3 +14,4 @@ TAGS
 /ligo/*.tz
 # We also ignore them in `haskell/test/resources/` because they are symlinked for tests.
 /haskell/test/resources/*.tz
+!/haskell/test/resources/transferlist.tz
```
gromakovsky, line 17: Since we decided to support only monolithic contract, it might be simpler to have ` /haskell/test/resources/stablecoin.tz` here instead of 2 lines.

docs/specification.md
```diff
@@ -92,6 +92,8 @@ Required `Safelist` entrypoints:
 * `assertReceivers(list address)`
```
gromakovsky, line 92: Note that apparently `assertReceiver` was removed, so we should remove it as well and use only `assertReceivers`.

haskell/nettest/Nettest.hs
```diff
@@ -23,10 +23,10 @@ scTransferScenario
      (Monad m, capsM ~ NettestT m)
   => (OriginationParams -> Storage)
   -> U.Contract
-  -> U.Contract -- Note that we do not distinguish safelist and stablecoin contracts
+  -> U.Contract -- Note that we do not distinguish transferlist and stablecoin contracts
   -> NettestImpl m
   -> m ()
-scTransferScenario constructInitialStorage stablecoinContract safelistContract = uncapsNettest $ do
+scTransferScenario constructInitialStorage stablecoinContract transferlistContract = uncapsNettest $ do
```
gromakovsky, line 29: Is it possible to update this scenario so that instead of transferlist contract it takes `originateTransferlist :: Set (Address, Address) -> Set Address -> m ()`? I. e. a callback that originates transferlist.

The only difficulty here is to construct external transferlist storage value from `Set (Address, Address)` (pairs with allowed transfers) and `Set Address` (permitted receivers). We can make this callback fail when someone tries to originate a transferlist with inconsistent data (e. g. AFAIU it's impossible to permit transfer from Alice to Bob and make `assertReceiver` fail for Bob). In that case we may need to update data we use `initialSafelistStorage`.

gromakovsky: Btw, I think `scTransferScenario` should be renamed because it tests more than just `transfer` now.

haskell/nettest/TransferlistIntegration.hs
```diff
@@ -0,0 +1,147 @@
+-- SPDX-FileCopyrightText: 2020 tqtezos
+-- SPDX-License-Identifier: MIT
+
+module TransferlistIntegration
+  ( scTransferlistIntegrationScenario
```
gromakovsky, line 5: If you manage to generalize `scTransferScenario`, this whole module should become unnecessary.

haskell/src/Indigo/Contracts/Transferlist/External.hs
```diff
@@ -0,0 +1,157 @@
...
+  = AssertTransfers [TransferParams]
+  | AssertReceivers [Address]
+  | OtherParameter ManagementParameter
+  deriving stock (Generic, Show)
+  deriving anyclass IsoValue
+
+instance ParameterHasEntryPoints Parameter where
+  type ParameterEntryPointsDerivation Parameter = EpdRecursive
+
+data ManagementParameter
```
gromakovsky, line 79: We don't want to call any entrypoints of transferlist except for `AssertTransfers` and `AssertReceivers`, so we don't need this type.

.buildkite/pipeline.yml
```diff
@@ -45,8 +45,10 @@ steps:
      MONEYBAG: "unencrypted:edsk3GjD83F7oj2LrnRGYQer99Fj69U2QLyjWGiJ4UoBZNQwS38J4v"
    commands: &run-nettest
      - nix build -L -f. nettest
+     - cd haskell/
      - eval "$FETCH_CONTRACT"
-     - nix run -f. tezos-client -c ./scripts/ci/run-local-chain-nettest.sh result/bin/stablecoin-nettest
+     - nix run -f ../ tezos-client -c ../scripts/ci/run-local-chain-nettest.sh ../result/bin/stablecoin-nettest
+     - cd ../
```
gromakovsky, line 51: Is it necessary? Does the next job start where this one stops?

haskell/package.yaml
```diff
@@ -21,6 +21,10 @@ verbatim:
 library:
   <<: *lib-common
 
+  ghc-options:
+    - -Wall
```
gromakovsky, line 25: `-Wall` here is useless because all warnings enabled by `-Wall` are already enabled, see `hpack/definitions.yaml`. Having `-Werror` in `.cabal` file is a bad practice because if someone else uses our library as a dependency and tries to build it with a different compiler version, they may be unable to do it because there can be new warnings in the new compiler version. What we actually want to do is to treat warnings as errors on CI, that should be done as part of CI configuration. If you need an example see https://github.com/serokell/tezos-globacap/blob/master/ci.nix

docs/specification.md
```diff
@@ -122,7 +122,7 @@ The next group consists of the errors that are not part of the FA2 specification
 | `ALLOWANCE_MISMATCH`         | In `configure_minter` both allowances are not `None`, but different                                                            |
 | `ADDR_NOT_MINTER`            | An attempt is made to modify minter data of an address that's not a minter                                                     |
 | `ALLOWANCE_EXCEEDED`         | Throws when trying to mint tokens more than currently allowed for an address                                                   |
-| `BAD_SAFELIST`               | Given address is a not a smart contract complying with the safelist interface                                                  |
+| `BAD_TRANSFERLIST`               | Given address is a not a smart contract complying with the transferlist interface                                                  |
```
gromakovsky, line 125:
```suggestion
| `BAD_TRANSFERLIST`           | Given address is a not a smart contract complying with the transferlist interface                                                  |
```

docs/specification.md
```diff
@@ -590,7 +590,7 @@ Parameter (in Michelson): `list nat`.
 
 - Burning tokens will not increase the mintingAllowance of the address doing the burning.
 
-- If safelist contract is set, this entrypoint MUST call its `assertReceiver` entrypoint (either directly or from the transfer hook) checking the minter (`SENDER`).
+- If transferlist contract is set, this entrypoint MUST call its `assertReceiver` entrypoint (either directly or from the transfer hook) checking the minter (`SENDER`).
```
gromakovsky, line 593:
```suggestion
- If transferlist contract is set, this entrypoint MUST call its `assertReceivers` entrypoint (either directly or from the transfer hook) checking the minter (`SENDER`).
```

haskell/test-common/Lorentz/Contracts/Test/Common.hs
```diff
@@ -62,6 +63,8 @@ commonOperator = genesisAddress6
 testOwner = genesisAddresses !! 7
 testPauser = genesisAddresses !! 8
 testMasterMinter = genesisAddresses !! 9
+testIssuer = testOwner
+testAdmin = testOwner
```
gromakovsky, line 67: Apparently `testIssuer` and `testAdmin` are used only in `Main`, so I wouldn't put them into `Common`.

haskell/src/Indigo/Contracts/Transferlist/External.hs
```diff
@@ -0,0 +1,77 @@
+-- SPDX-FileCopyrightText: 2020 Globacap
```
gromakovsky, line 1: It's actually a bit weird. Did you write `Globacap` here because you think you copy-paste a substantial copy of https://github.com/serokell/tezos-globacap/blob/master/nettest/Nettest/WhitelistIntegration.hs? I think only `Storage` definitions sort of resembles `WhitelistStorage`, the rest looks like a new code.

haskell/nettest/Nettest.hs
```diff
@@ -18,15 +17,16 @@ import Util.Named
...
 import Lorentz.Contracts.Stablecoin
 
-scTransferScenario
+scNettestScenario
   :: forall m capsM.
      (Monad m, capsM ~ NettestT m)
   => (OriginationParams -> Storage)
   -> U.Contract
-  -> U.Contract -- Note that we do not distinguish safelist and stablecoin contracts
+  -- -> U.Contract -- Note that we do not distinguish transferlist and stablecoin contracts
```
gromakovsky, line 25: Looks like a leftover.

haskell/test/resources/transferlist.tz
```diff
@@ -0,0 +1,322 @@
+# SPDX-FileCopyrightText: 2020 Michael J. Klein Tocqueville Group
+# SPDX-License-Identifier: MIT
```
gromakovsky, line 2: `LICENSES/MIT.txt` is not the same file as https://github.com/tqtezos/lorentz-contract-transferlist/blob/master/LICENSE. You should add `LICENSES/LicenseRef-MIT-Klein.txt` and refer to it here.

gromakovsky:
Also please update the changelog and fix a formatting issue present there (`[#73] Update the contract to a newer FA2 revision:` → `[#73](Update the contract to a newer FA2 revision):`).

docs/specification.md
```diff
@@ -79,19 +79,19 @@ Stablecoin contract MUST provide an additional metadata `big_map` for its tokens
...
+Required `Transferlist` entrypoints:
 * `assertTransfers(list (pair (address %from) (list %tos address)))`
   * Checks whether a transfer is permitted from one address to the other one.
   Fails if transfer is prohibited for any item from this list.
-* `assertReceiver(address)`
-  * Fails if address is not whitelisted or it is blacklisted.
 * `assertReceivers(list address)`
   * Fails if any address in the list is not whitelisted or is blacklisted.
 
+The specific interface of transferlist contracts is described in [TZIP-15](https://gitlab.com/tzip/tzip/-/blob/c38f4c2dc9da883d01e85409d34d0015d6ad272a/proposals/tzip-15/tzip-15.md)
```
sras, line 93: It might be worth adding that we need `UserId` to be an `Address` for the Transferlist contract to be compatible with this contract.

ligo/stablecoin/spec.ligo
```diff
@@ -253,7 +253,7 @@ type closed_parameter is
 | Accept_ownership      of accept_ownership_param
 | Change_master_minter  of change_master_minter_param
 | Change_pauser         of change_pauser_param
-| Set_safelist          of set_safelist_param
+| Set_transferlist          of set_transferlist_param
```
gromakovsky, line 256:
```suggestion
| Set_transferlist      of set_transferlist_param
```

awkure: done

haskell/test/resources/transferlist.tz
awkure: done

haskell/nettest/Nettest.hs
awkure: done

haskell/src/Indigo/Contracts/Transferlist/External.hs
awkure: done

haskell/src/Indigo/Contracts/Transferlist/External.hs
```diff
@@ -0,0 +1,79 @@
...
+data Storage = Storage
+  { issuer :: Address
+  , users  :: BigMap Address TransferlistId
+  , transferlists :: BigMap TransferlistId OutboundTransferlists
+  , admin :: Address
+  }
+  deriving stock Generic
+  deriving anyclass (IsoValue, HasTypeAnn)
+
+{-# ANN convertToExternalStorage ("HLint: ignore Use 'head' from Universum.Unsafe" :: Text) #-}
```
gromakovsky, line 39: Why don't you want to use `Unsafe.head` here btw?

ChangeLog.md
```diff
@@ -6,7 +6,7 @@ SPDX-License-Identifier: MIT
 
 ## 1.0.0
 
-* [#73] Update the contract to a newer FA2 revision:
+* [#73] (Update the contract to a newer FA2 revision):
```
gromakovsky, line 9:
```suggestion
* [#73](Update the contract to a newer FA2 revision):
```

AFAIU there should be no space.

gromakovsky:
By "update the changelog" I meant adding a new entry there. I think we can make 1.1.0 release right after this PR, so you can call it 1.1.0 straight away.

gromakovsky:
Commits are quite messed up btw.

haskell/src/Indigo/Contracts/Transferlist/External.hs
awkure: This leads adding `universum` to dependencies

gromakovsky: I mean `Unsafe` from `morley-prelude`, not `Universum.Unsafe` from `universum`: https://hackage.haskell.org/package/morley-prelude-0.3.0/docs/Unsafe.html

That being said, maybe `hlint` will trigger this warning for `Unsafe` because it's not exactly `Universum.Unsafe`, but I'd try at least.

Issue #77, opened 2020-07-20
Implement stablecoin-client


gromakovsky:
# Clarification and motivation

We would like to make it easy to use the stablecoin contract. Currently this repository offers the smart contract itself and a Haskell library that can be used by developers. End users have to use `ligo` to compile the contract to Michelson and construct the initial storage and `tezos-client` (or similar software) to interact with it. The workflow can be made more convenient. One possible way is to develop a command line client tool to interact with the contract.

We have already done it in [tezos-btc](https://github.com/tz-wrapped/tezos-btc). Let's do something similar for `stablecoin`.

# Acceptance criteria

There is a new executable `stablecoin-client` that can:
1. Deploy `stablecoin` contract. In `stablecoin` deployment consists of just a single origination operation (unlike tzbtc).
2. Call any entrypoint using a specific interface (1 CLI command per entrypoint, read-only entrypoints are considered below).
3. Read any data from the contract (get balance, metadata, etc.).

Let's use `morley-client` (library) for that. Long ago we implemented client functionality in `tezos-btc`, then we ported it to `morley-client`, so now it's duplicated between those two repos. Let's not copy-paste it again.

Note that in `tezos-btc` we have nettest-based tests that use `tzbtc-client` under the hood. They were added [here](https://github.com/tz-wrapped/tezos-btc/pull/108). It'd be nice to have such tests for `stablecoin-client` as well and run them on CI.

Issue #53, opened 2020-05-27
Describe errors in the specification


gromakovsky:
# Clarification and motivation

It would be nice to specify what error messages our contract can fail with in the specification, so that error scenarios are clearer to the users.

# Acceptance criteria

For each entrypoint we specify errors this entrypoint can fail with. We also say that there are internal errors that should never happen as long as the contract is originated properly.

gromakovsky:
I will make the table more accurate and update each entrypoint's description with error scenarios after #35, #41, #54, #55, #56.

PR #79, opened 2020-07-27
[#77] Implement stablecoin-client

dcastro:
## Description

This PR introduces a new `stablecoin-executable`.

This executable supports commands for deploying the stablecoin contract, calling its endpoints and querying its state.

Query commands that need to inspect the storage's `big_map`s will be introduced in a separate PR, as those require some additional changes in the `morley-client` library upstream.

## Related issue(s)

Resolves part of #77

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/haskell/README.md
+++ b/haskell/README.md
@@ -5,12 +5,28 @@ SPDX-License-Identifier: MIT
 
 # Stablecoin Haskell
 
-This folder contains a Haskell library to interact with the stablecoin contract (basically Haskell bindings) and tests for it.
+This folder contains:
+
+1. a Haskell library to interact with the stablecoin contract (basically Haskell bindings) and tests for it.
+2. a `stablecoin-client` executable to deploy and interact with the stablecoin contract.
 
 ## Build Instructions
 
 You need [Stack](http://haskellstack.org/) to build this package.
-Run `stack build` to build the library.
+
+To build the library and the executable:
+1. Copy the stablecoin contract to `test/resources/stablecoin.tz` and then run `stack build`.
+2. Or, alternatively, run `make build`.
+
+Run `stack install` to install the `stablecoin-client` globally.
+
+## Run Instructions
+
+Pre-requisites: `tezos-client` must be installed.
+
+If the `stablecoin-client` was installed globally with `stack install`, then run `stablecoin-client --help`
+to see a list of the available commands.
+Otherwise, run `stack run stablecoin-client -- --help`.
 
 ## Tests
 
--- a/haskell/nettest/Main.hs
+++ b/haskell/nettest/Main.hs
@@ -11,7 +11,7 @@ import qualified Indigo.Contracts.Transferlist.External as External
 import qualified Indigo.Contracts.Transferlist.Internal as Internal
 import Lorentz (ToT, compileLorentzContract, toVal)
 import Lorentz.Contracts.Test.Common
-import Michelson.Runtime (parseExpandContract, prepareContract)
+import Michelson.Runtime (parseExpandContract)
 import Michelson.Typed (untypeValue)
 import Michelson.Typed.Convert (convertContract)
 import Morley.Nettest
@@ -19,6 +19,8 @@ import Nettest (scNettestScenario)
 import Tezos.Address
 import Util.IO
 import Util.Named
+
+import Stablecoin.Client.Contract (parseStablecoinContract)
 
 externalTransferlistContractPath :: FilePath
 externalTransferlistContractPath = "test/resources/transferlist.tz"
@@ -32,7 +34,7 @@ main = do
       (#description .! "Stablecoin nettest scenarioWithInternalTransferlist")
       (#header .! "ManagedLedger nettest")
       (#parser .! clientParser)
-  stablecoinContract <- prepareContract $ Just "test/resources/stablecoin.tz"
+  stablecoinContract <- parseStablecoinContract
   externalTransferlistFile <- readFileUtf8 externalTransferlistContractPath
   externalTransferlistContract <- either (error "cannot parse transferlist contract") pure $
     parseExpandContract (Just externalTransferlistContractPath) externalTransferlistFile
--- a/haskell/nettest/Main.hs
+++ b/haskell/nettest/Main.hs
@@ -15,19 +15,22 @@ import Michelson.Runtime (parseExpandContract)
 import Michelson.Typed (untypeValue)
 import Michelson.Typed.Convert (convertContract)
 import Morley.Nettest
-import Nettest (scNettestScenario)
 import Tezos.Address
 import Util.IO
 import Util.Named
 
+import Nettest (scNettestScenario)
+import Stablecoin.Client.Cleveland.Caps (runStablecoinClient)
 import Stablecoin.Client.Contract (parseStablecoinContract)
+import StablecoinClientTest (stablecoinClientScenario)
 
 externalTransferlistContractPath :: FilePath
 externalTransferlistContractPath = "test/resources/transferlist.tz"
 
 main :: IO ()
 main = do
-  let clientParser = clientConfigParser . pure $ Just "nettest.Stablecoin"
+  let aliasPrefix = "nettest.Stablecoin"
+  let clientParser = clientConfigParser . pure $ Just aliasPrefix
   parsedConfig <- execParser $
     parserInfo
       (#usage .! mempty)
@@ -75,3 +78,5 @@ main = do
 
   runNettestClient env scenarioWithInternalTransferlist
   runNettestClient env scenarioWithExternalTransferlist
+
+  runStablecoinClient parsedConfig env (stablecoinClientScenario aliasPrefix)
--- a/haskell/nettest/Nettest.hs
+++ b/haskell/nettest/Nettest.hs
@@ -90,7 +90,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
       (#amount .! value)
 
     callTransferWithOperator op from to value =
-      callFrom (AddressResolved op) sc (ep "transfer") (tp from to value)
+      callFrom' (AddressResolved op) sc (ep "transfer") (tp from to value)
 
     callTransfer = join callTransferWithOperator
 
@@ -98,7 +98,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
       :: [("from_" :! Address, [("to_" :! Address, "amount" :! Natural)])]
       -> capsM ()
     callTransfers = mapM_ $ \(from@(arg #from_ -> from_), destinations) ->
-      callFrom
+      callFrom'
         (AddressResolved from_)
         sc
         (ep "transfer")
@@ -106,7 +106,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
 
     configureMinter :: Address -> Address -> Maybe Natural -> Natural -> capsM ()
     configureMinter from for expectedAllowance newAllowance =
-      callFrom
+      callFrom'
         (AddressResolved from)
         sc
         (ep "configure_minter")
@@ -116,7 +116,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
 
     removeMinter :: Address -> Address -> capsM ()
     removeMinter from whom =
-      callFrom
+      callFrom'
         (AddressResolved from)
         sc
         (ep "remove_minter")
@@ -124,7 +124,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
 
     addOperatorNettest :: Address -> Address -> capsM ()
     addOperatorNettest from op =
-      callFrom
+      callFrom'
         (AddressResolved from)
         sc
         (ep "update_operators")
@@ -132,7 +132,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
 
     removeOperator :: Address -> Address -> capsM ()
     removeOperator from op =
-      callFrom
+      callFrom'
         (AddressResolved from)
         sc
         (ep "update_operators")
@@ -140,7 +140,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
 
     mint :: Address -> Natural -> capsM ()
     mint to_ value =
-      callFrom
+      callFrom'
         (AddressResolved to_)
         sc
         (ep "mint")
@@ -148,7 +148,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
 
     burn :: Address -> Natural -> capsM ()
     burn from amount_ =
-      callFrom
+      callFrom'
         (AddressResolved from)
         sc
         (ep "burn")
@@ -156,7 +156,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
 
     pause :: Address -> capsM ()
     pause from =
-      callFrom
+      callFrom'
         (AddressResolved from)
         sc
         (ep "pause")
@@ -164,7 +164,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
 
     unpause :: Address -> capsM ()
     unpause from =
-      callFrom
+      callFrom'
         (AddressResolved from)
         sc
         (ep "unpause")
@@ -172,7 +172,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
 
     transferOwnership :: Address -> Address -> capsM ()
     transferOwnership from to =
-      callFrom
+      callFrom'
         (AddressResolved from)
         sc
         (ep "transfer_ownership")
@@ -180,7 +180,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
 
     acceptOwnership :: Address -> capsM ()
     acceptOwnership from =
-      callFrom
+      callFrom'
         (AddressResolved from)
         sc
         (ep "accept_ownership")
@@ -188,7 +188,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
 
     changeMasterMinter :: Address -> Address -> capsM ()
     changeMasterMinter from to =
-      callFrom
+      callFrom'
         (AddressResolved from)
         sc
         (ep "change_master_minter")
@@ -196,7 +196,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
 
     changePauser :: Address -> Address -> capsM ()
     changePauser from to =
-      callFrom
+      callFrom'
         (AddressResolved from)
         sc
         (ep "change_pauser")
@@ -204,7 +204,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
 
     setTransferlist :: Address -> Address -> capsM ()
     setTransferlist from transferlistAddress =
-      callFrom
+      callFrom'
         (AddressResolved from)
         sc
         (ep "set_transferlist")
@@ -212,7 +212,7 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
 
     unsetTransferlist :: Address -> capsM ()
     unsetTransferlist from =
-      callFrom
+      callFrom'
         (AddressResolved from)
         sc
         (ep "set_transferlist")
@@ -344,3 +344,29 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
   burnScenario
   permissionReassigmentScenario
   transferlistScenario
+
+-- This is a temporary solution.
+-- Changes made in cleveland's `callFrom` have caused some issues when calling
+-- entrypoints other than the default.
+-- So we've temporarily copied the old `callFrom` implementation here and
+-- renamed it to `callFrom'`.
+-- See discussion here: https://github.com/tqtezos/stablecoin/pull/79#discussion_r464983707
+--
+-- TODO: delete this after these comments have been addressed:
+-- https://gitlab.com/morley-framework/morley/-/merge_requests/499#note_388448660
+callFrom'
+  :: (MonadNettest caps base m, NiceParameter v)
+  => AddressOrAlias
+  -> AddressOrAlias
+  -> EpName
+  -> v
+  -> m ()
+callFrom' from to epName param =
+  transfer $
+    TransferData
+      { tdFrom = from
+      , tdTo = to
+      , tdAmount = zeroMutez
+      , tdEntrypoint = epName
+      , tdParameter = param
+      }
--- a/haskell/src/Stablecoin/Client/Cleveland/Caps.hs
+++ b/haskell/src/Stablecoin/Client/Cleveland/Caps.hs
@@ -1,0 +1,259 @@
+-- SPDX-FileCopyrightText: 2020 tqtezos
+-- SPDX-License-Identifier: MIT
+
+{-# LANGUAGE InstanceSigs #-}
+
+module Stablecoin.Client.Cleveland.Caps
+  ( StablecoinScenario
+  , MonadStablecoin
+  , runStablecoinClient
+  , stablecoinCapImpl
+  -- * Operations
+  , deploy
+  , transfer
+  , getBalanceOf
+  , updateOperators
+  , isOperator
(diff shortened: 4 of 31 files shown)


haskell/src/Client/Contract.hs
```diff
@@ -0,0 +1,73 @@
...
+stablecoinContractContent = $(embedStringFile "../ligo/stablecoin.tz")
+
+parseStablecoinContract :: MonadThrow m => m U.Contract
+parseStablecoinContract =
+  either throwM pure $
+    parseExpandContract
+      (Just "../ligo/stablecoin.tz")
+      stablecoinContractContent
+
+mkInitialStorage :: Address -> Address -> Address -> Address -> MText -> MText -> Natural -> Storage
```
gromakovsky: I don't like that there are so many arguments and some of them have same types. It looks similar to `DeploymentData`, but there we have `AddressOrAlias` and here we have `Address`. What about parameterizing `DeploymentData` with `a` and passing `DeploymentData Address` here (and `DeploymentData AddressOrAlias` in another place)?

haskell/src/Client/StablecoinClient/Class.hs
```diff
@@ -0,0 +1,138 @@
...
+  { ddMasterMinter :: AddressOrAlias
+  , ddContractOwner :: AddressOrAlias
+  , ddPauser :: AddressOrAlias
+  , ddSafelist :: AddressOrAlias
+  , ddTokenName :: MText
+  , ddTokenSymbol :: MText
+  , ddTokenDecimals :: Natural
+  }
+
+class Monad m => HasStablecoinClient m where
```
gromakovsky, line 46: What is this type class needed for? AFAIU, right now it has only one instance defined right here. Wouldn't it be simpler to have `deploy ::  "sender" :! AddressOrAlias -> DeploymentData -> MorleyClientM Address`? Are you going to add another implementation for tests? That's one possible approach (to mock these methods), but we can have tests for real implementation where we don't mock anything and these tests are more useful. We are successfully doing it in `tezos-btc`. There we started with mock tests because setting up and running tests for real implementation was complicated. But now it's not complicated, we are successfully running nettests on CI. So usefulness of those mock tests seems to be small. And after we do morley's #53 we should be able to write tests using the same interface that will run in emulated environment, on real network using nettest implementation and on real network using `stablecoin-client`.

haskell/src/Client/Contract.hs
```diff
@@ -0,0 +1,73 @@
...
+import Lorentz.Contracts.Stablecoin (Storage, mkTokenMetadata)
+
+stablecoinContractContent :: Text
+stablecoinContractContent = $(embedStringFile "../ligo/stablecoin.tz")
+
+parseStablecoinContract :: MonadThrow m => m U.Contract
+parseStablecoinContract =
+  either throwM pure $
+    parseExpandContract
+      (Just "../ligo/stablecoin.tz")
```
gromakovsky: Since we are also using the contract in tests, I would define `stablecoinPath :: FilePath` in the library, export it and use in tests, so that the path does not appear in multiple places.

haskell/README.md
```diff
@@ -5,12 +5,23 @@ SPDX-License-Identifier: MIT
...
-Run `stack build` to build the library.
+
+Run `stack build` to build the library and the executable.
+Run `stack install` to install the `stablecoin-client` globally.
+
+## Run Instructions
+
+If the `stablecoin-client` was installed globally with `stack install`, then run `stablecoin-client --help`
+to see a list of the available commands.
+Otherwise, run `stack run stablecoin-client -- --help`.
```
gromakovsky, line 24: I'd add that `tezos-client` must be installed in order to use `stablecoin-client`.

haskell/src/Client/StablecoinClient/Class.hs
dcastro: Ah good point! Yeah, I can't think of a scenario where you would run deploy on anything other than `MorleyClientM`, so I'll get rid of the typeclass and simplify this.

haskell/src/Client/Contract.hs
dcastro: Done

haskell/README.md
dcastro: Done

haskell/src/Client/Contract.hs
dcastro: I like that! Done.

haskell/src/Client/StablecoinClient/Class.hs
dcastro: Done.

Also moved everything under `Stablecoin.*` (e.g. `Stablecoin.Client.Parser`) to mirror the module hierarchy we used in `Morley.Client`.

balsoft:
I'd prefer to have the whole contract business to happen inside the nix expressions, so that `nix build` in a fresh checkout actually works.

haskell/src/Stablecoin/Client/Impl.hs
```diff
@@ -0,0 +1,322 @@
...
+        , isdPauser = pauser
+        , isdTransferlist = transferlist
+        }
+
+  let initialStorage = mkInitialStorage initialStorageData'
+
+  originateUntypedContract
+    False
+    alias
+    (AddressResolved senderAddr)
```
gromakovsky, line 81: It seems you can pass `sender` here and avoid doing `resolveAddress sender`. It's the same logic, but a bit shorter.

haskell/src/Stablecoin/Client/Impl.hs
```diff
@@ -0,0 +1,322 @@
...
+    UnknownAddress _ -> pure Nothing
+    err -> throwM err
+  pure $ AddressAndAlias addr aliasMb
+
+-- | Call an entrypoint of the stablecoin contract with the given argument.
+call
+  :: forall epRef epArg.
+     HasEntryPointArg Parameter epRef epArg
+  => IsoValue epArg
+  => Typeable epArg
```
gromakovsky, line 237: It looks somewhat unusual to write constraints as `a => b => c` instead of `(a, b) => c`. Do you prefer this style or does it have any advantages?

haskell/stack.yaml
```diff
@@ -30,13 +30,6 @@ extra-deps:
 - show-type-0.1.1
 - git: https://github.com/int-index/caps.git
   commit: c5d61837eb358989b581ed82b1e79158c4823b1b
-# tezos-bake-monitor-lib and its deps which are used in morley-client
```
gromakovsky, line 33: It seems there are some dependencies of `tezos-bake-monitor-lib` below that are not needed anymore. Could you please check whether there are useless extra-deps and delete them?

haskell/src/Stablecoin/Client/Impl.hs
dcastro: No particular advantages, though I personally like the similarity to `a -> b -> c` :) Should I change it for consistency?

haskell/src/Stablecoin/Client/Impl.hs
dcastro: Done

dcastro:
@gromak I've uploaded a first rough draft of how I'm implementing the nettest functionality.

It's slightly different from `tezos-btc`, because that approach (e.g. re-implementing `originateUntyped` in terms of `tzbtc deploy`, `transfer` in terms of `tzbtc <entrypoint name>`) is a bit limited.

While that approach works for invoking `deploy` and client commands that call entrypoints, I don't think it works for other commands like `--get-paused` or `--get-master-minter`.

So my strategy was to create a new capability `StablecoinImpl` (which can be used alongside the existing `NettestImpl` capability).

It's still WIP, I'll upload the rest asap.

haskell/stack.yaml
dcastro: Done

gromakovsky:
Nitpick: can you please wrap this text?
image

dcastro:
@gromakovsky Done!

haskell/src/Stablecoin/Client/Cleveland/Caps.hs
```diff
@@ -0,0 +1,229 @@
...
+-- TODO: this already exists in Cleveland, but it's not exported.
+-- We should export it.
+revealKeyUnlessRevealed :: MonadStablecoin caps base m => Alias -> m ()
+revealKeyUnlessRevealed = actionToCaps siRevealKeyUnlessRevealed
+
+----------------------------------------------------------------------------
+-- Internal helpers
+----------------------------------------------------------------------------
+
+-- TODO: this already exists in Cleveland, but it's not exported.
```
gromakovsky, line 216: Regarding `caps`, I have just created an issue I wanted to create long ago: https://gitlab.com/morley-framework/morley/-/issues/309. No need to change anything here, just sharing some thoughts.

gromakovsky:
Some files are :so-big: and I haven't reviewed them carefully yet in this round of review, but I plan to review them by the end of the day (need to take some break). Everything else looks good.

dcastro:
> Some files are :so-big:

And they're about to get bigger because I just realized I completely missed the FA2 entrypoints, so I'm gonna add those too now :sweat_smile:

haskell/src/Stablecoin/Client/Parser.hs
```diff
@@ -0,0 +1,522 @@
...
+
+naturalOption ::
+  Maybe Natural -> "name" :! String -> "help" :! String -> Opt.Parser Natural
+naturalOption = mkCLOptionParser
+
+nonEmptyParser :: Opt.Parser a -> Opt.Parser (NonEmpty a)
+nonEmptyParser p =
+  -- Note: `(:|) <$> p <*> many p` does not work here, because then optparse-applicative
+  -- will display the option's info twice when --help is invoked.
+  -- For this reason, we have to use `some` and the "unsafe" `NonEmpty.fromList`.
```
gromakovsky, line 521: Interesting. It might be a good idea to upstream this function to `optparse-applicative` some time later.

gromakovsky:
Okay, I've made another pass, couldn't find anything wrong. Will make one more when remaining entrypoints are added.

dcastro:
K, I've pushed the FA2 entrypoints + nettests.
Now I just need to ensure `stablecoin-client` is installed before running the tests in the pipeline.

haskell/nettest/Nettest.hs
```diff
@@ -80,17 +81,14 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
...
-
     tp :: Address -> Address -> Natural -> FA2.TransferParams
     tp from to value = constructSingleTransfer
       (#from_ .! from)
       (#to_ .! to)
       (#amount .! value)
 
     callTransferWithOperator op from to value =
-      callFrom (AddressResolved op) sc (ep "transfer") (tp from to value)
+      callFrom (AddressResolved op) sc DefEpName (Call_FA2 . FA2.Transfer $ tp from to value)
```
gromakovsky, line 91: It may be out of scope of this PR, but I like the previous version more. There `tezos-client` and RPC would be called with a small argument value and explicit entrypoint name, while in this version we always pass a large value wrapped into `Right` and `Left` constructors. So the previous version at least has cleaner logs. I am not sure whether there will be difference in block explorers, i. e. whether they can realize that `transfer` is called when you call the default entrypoint and pass `Left (Right (blabla…`. Anyway, I think the previous version should be preferred at least for cleaner logs.

AFAIU it's done this way because of [this](https://gitlab.com/morley-framework/morley/-/merge_requests/499#note_388408872). So probably the problem is out of scope of this PR and what we have now is the only possible fast solution.

haskell/src/Stablecoin/Client/Impl.hs
```diff
@@ -261,14 +338,27 @@ getStorage (arg #contract -> contract) = do
...
     Left err -> throwM $ SCEExpressionParseError storageExpr err
 
+-- | Safer version of 'Client.readBigMapValue' that accepts a typed `BigMapId`.
+safeReadBigMapValue
+  :: (NicePackedValue k, NiceUnpackedValue v, UnpackedValScope (ToT v))
+  => BigMapId k v -> k -> MorleyClientM v
+safeReadBigMapValue (BigMapId bigMapId) k = Client.readBigMapValue bigMapId k
+
+-- | Safer version of 'Client.readBigMapValueMaybe' that accepts a typed `BigMapId`.
+safeReadBigMapValueMaybe
```
gromakovsky, line 348: Wouldn't it be better to put it into `morley` repository (along with typed `BigMapId`)?

haskell/stack.yaml
```diff
@@ -11,11 +11,11 @@ extra-deps:
     https://gitlab.com/morley-framework/morley.git
     # ^ CI cannot use ssh, so we use http clone here
   commit:
-    37449aa1c4e889557058e6895822fa97e5a0215a # master
+    895a659d9faf2f32ecdd11fe39ee8edd4d09146f # master
```
gromakovsky, line 14: Just in case: https://gitlab.com/morley-framework/morley/-/merge_requests/532 should be merged first and this reference should be updated to actually point to master.

haskell/src/Stablecoin/Client/Parser.hs
```diff
@@ -0,0 +1,684 @@
...
+  }
+  deriving stock Show
+
+data GlobalOptions = GlobalOptions
+  { goUser :: AddressOrAlias
+  , goContract :: AddressOrAlias
+  }
+  deriving stock Show
+
+-- TODO: permissions descriptor
```
gromakovsky, line 55: Looks like an old TODO.

haskell/src/Stablecoin/Client/Impl.hs
dcastro: Done

haskell/stack.yaml
dcastro: Yup, I'll try to get !532 reviewed and merged asap.

haskell/nettest/Nettest.hs
dcastro: Agreed, the previous version looks a lot better. I'll make a note to come back to this after addressing the comments in morley's !499.

haskell/src/Stablecoin/Client/Parser.hs
dcastro: Good catch! Removed.

haskell/stack.yaml
dcastro: On that note, do you know if that's the reason why the [pipeline build is failing](https://buildkite.com/serokell/stablecoin/builds/735#7e0e785e-0b91-4eb7-81a0-a07bfcbf1445/63-90)? 

It's weird that `stack` builds this just fine locally, regardless of whether the commit exists in master or not, but it fails in the pipeline...

gromakovsky: Oh, yes, that's a known problem previously discussed in this thread: https://serokell.slack.com/archives/CUMAQ3PL4/p1592934064227100. And it looks like the problem is quite deep :(

haskell/src/Stablecoin/Client/Parser.hs
```diff
@@ -0,0 +1,683 @@
...
+      , "EXAMPLE:", linebreak
+      , "  stablecoin-client mint --help", linebreak
+      , "USAGE EXAMPLE:", linebreak
+      , "  stablecoin-client mint --to tz1U1h1YzBJixXmaTgpwDpZnbrYHX3fMSpvb --amount 100500", linebreak
+      , linebreak
+      , "  This command will produce the given amounts of tokens to", linebreak
+      , "  the wallets associated with the given address.", linebreak
+      ]
+
+-- TODO: add this orphan instance to morley
```
gromakovsky, line 661: Maybe let's add it in https://gitlab.com/morley-framework/morley/-/merge_requests/532?

haskell/src/Stablecoin/Client/Cleveland/IO.hs
```diff
@@ -0,0 +1,166 @@
...
+  (fmap unsafeMkMutez decimal <* string " μꜩ")
+
+naturalParser :: Parser Natural
+naturalParser = decimal
+
+textParser :: Parser Text
+textParser = fromString <$> some printChar
+
+mTextParser :: Parser MText
+mTextParser = either error id . mkMText <$> textParser
```
gromakovsky, line 118: I think it's better to fail using some `megaparsec`'s function rather than `error`. Though, since it's used only in tests and only to parse the output of `stablecoin-client`, I think `error` is acceptable as well, but then please add `HasCallStack` just in case.

haskell/src/Stablecoin/Client/Parser.hs
```diff
@@ -0,0 +1,683 @@
...
+  | CmdGetTotalSupply
+  | CmdGetMintingAllowance GetMintingAllowanceOptions
+  | CmdGetTokenMetadata
+  deriving stock Show
+
+data DeployContractOptions = DeployContractOptions
+  { dcoMasterMinter :: AddressOrAlias
+  , dcoContractOwner :: AddressOrAlias
+  , dcoPauser :: AddressOrAlias
+  , dcoTransferlist :: AddressOrAlias
```
gromakovsky, line 92: Why is transferlist not optional? It's optional in the contract's storage.

haskell/src/Stablecoin/Client/Parser.hs
```diff
@@ -0,0 +1,683 @@
...
+        mTextOption
+          Nothing
+          (#name .! "token-name")
+          (#help .! "Token name")
+      dcoTokenDecimals <-
+        naturalOption
+          Nothing
+          (#name .! "token-decimals")
+          (#help .! ("Number of digits to use after the decimal point " <>
+                     "when displaying the token amounts"))
```
gromakovsky, line 264: I think token metadata will often be out of interest, so I would define some default values for `dcoTokenSymbol`, `dcoTokenName`, `dcoTokenDecimals`.

haskell/nettest/Nettest.hs
```diff
@@ -344,3 +345,20 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
   burnScenario
   permissionReassigmentScenario
   transferlistScenario
+
+callFrom'
```
gromakovsky, line 349: I'd add a comment here explaining why this function is necessary and saying that it may become unnecessary because we may reconsider upstream `callFrom`.

haskell/nettest/Nettest.hs
```diff
@@ -344,3 +345,20 @@ scNettestScenario constructInitialStorage stablecoinContract originateTransferli
...
+  -> AddressOrAlias
+  -> EpName
+  -> v
+  -> m ()
+callFrom' from to epName param =
+  transfer $
+    TransferData
+      { tdFrom = from
+      , tdTo = to
+      , tdAmount = unsafeMkMutez 0
```
gromakovsky, line 361: Btw, now there is `zeroMutez`.

haskell/src/Stablecoin/Client/Parser.hs
dcastro: Done

haskell/stack.yaml
```diff
@@ -1,21 +1,22 @@
 # SPDX-FileCopyrightText: 2020 tqtezos
 # SPDX-License-Identifier: MIT
 
-resolver: lts-15.7
+resolver: lts-16.5
 
 packages:
 - .
 
+# e2d3cf9197804a2cfa92715e1a4d640fe5b87af7 # master
```
gromakovsky, line 9: Is it going to be removed?

haskell/nettest/Main.hs
```diff
@@ -10,29 +10,34 @@ import qualified Unsafe (fromJust)
...
+  let aliasPrefix = "nettest.Stablecoin"
+  let clientParser = clientConfigParser . pure $ Just aliasPrefix
   parsedConfig <- execParser $
     parserInfo
       (#usage .! mempty)
       (#description .! "Stablecoin nettest scenarioWithInternalTransferlist")
       (#header .! "ManagedLedger nettest")
       (#parser .! clientParser)
-  stablecoinContract <- prepareContract $ Just "test/resources/stablecoin.tz"
+  stablecoinContract <- prepareContract $ Just stablecoinPath
```
gromakovsky, line 40: Since the contract is now embedded into source code, don't we want to use the embedded version here? I. e. use `parseStablecoinContract`. I know they are supposed to be the same, but in theory one may compile code from one folder and run tests from another folder (even on another machine) in which case `stablecoin-client` will use not the same contract as some tests. Also if we use `parseStablecoinContract` here, it will be easier to run these tests because we won't have to make sure that `"test/resources/stablecoin.tz"` is available when we **run** the tests.

haskell/nettest/Nettest.hs
dcastro: FYI, I ended up reverting these changes. Apparently, calling `DefEpName` with `Call_FA2 (FA2.Transfer x)` is not equivalent to calling `ep "transfer"` with `x`. I was under the impression that they were equivalent, but it turns out I was getting [errors like this](https://buildkite.com/serokell/stablecoin/builds/737#e9d58043-8051-463d-a43f-65461f2a93b0/63-109):

```
src/Morley/Nettest/Pure.hs:52
  | Interpreter failed: The contract parameter is ill-typed: Error checking value Left { Pair "tz1f1S7V2hZJ3mhj47djb5j1saek8c2yB2Cx" { Pair "tz1f1S7V2hZJ3mhj47djb5j1saek8c2yB2Cx" (Pair 0 111) } } against type unit: unknown value Error occured on line 1 char 1.
```

I added `callFrom'` [here](https://github.com/tqtezos/stablecoin/pull/79/files#diff-80418a2f9a5a7631759638be01c0d856R348-R366) as a temporary measure. I'll come back to this later with a more permanent solution.

haskell/nettest/Main.hs
dcastro: Good idea! Done, and changed this in a couple of other places as well.

haskell/stack.yaml
dcastro: Done

haskell/nettest/Nettest.hs
dcastro: Thanks! Done

haskell/nettest/Nettest.hs
dcastro: Done

haskell/src/Stablecoin/Client/Cleveland/IO.hs
dcastro: Done

haskell/src/Stablecoin/Client/Parser.hs
dcastro: I was under the impression it was mandatory when I first began working on this issue, but now I don't think it needs to be anymore. I'll make this change.

haskell/src/Stablecoin/Client/Parser.hs
dcastro: I suppose [these](https://github.com/tqtezos/stablecoin/blob/master/haskell/src/Lorentz/Contracts/Stablecoin.hs#L236-L238) would qualify as sensible defaults, is that right?

```haskell
      (#token_id .! 0, #mdr .! ( #symbol .! [mt|USDC|]
                               , #mdr2 .! ( #name .! [mt|USDC|]
                                          , #mdr3 .! ( #decimals .! 8, #extras .! mempty))))
```

gromakovsky: Yes, that's fine.

dcastro: Done

haskell/src/Stablecoin/Client/Parser.hs
dcastro: Done

gromakovsky:
LGTM, the only remaining thing is to make CI pass :)

dcastro:
bliss

haskell/src/Stablecoin/Client/Parser.hs
dcastro: Seems like a solution has been very recently added to `optpatse-applicative`, and will be in the next release:

* https://github.com/pcapriotti/optparse-applicative/issues/376
* https://github.com/pcapriotti/optparse-applicative/pull/382

I'll update my comment here to mention this, and that we should use `some1` once it's released.

PR #83, opened 2020-08-05
[#78] Produce static stablecoin-client for linux on CI

### LINKED ISSUE #78: Produce static stablecoin-client for linux on CI

In #77 we are going to implement `stablecoin-client`. Users don't want to build it from sources, we need to make it more convenient to obtain it. Our typical approach is to build static executable with nix, for example we are doing it in `tezos-btc`.

# Acceptance criteria

CI artifacts contain fully static `stablecoin-client` executable that runs on any linux.

dcastro:
## Description

Problem: In #77, we added the `stablecoin-client` executable. Users
don't want to build it from sources, we need to make it more convenient
to obtain it. 

Solution: Configure CI to build an artifact with the static executable.

Example build with the `stablecoin-client` artifact: https://buildkite.com/serokell/stablecoin/builds/754#4cf20eff-41e8-4866-97c0-f0b9a07f1707

## Related issue(s)

Resolves #78

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -75,6 +75,14 @@ steps:
      # weeder needs .cabal file:
      - nix run -f.. pkgs.haskellPackages.hpack -c hpack
      - ./result
+ - label: packaging
+   if: *not_scheduled
+   commands:
+     - eval "$FETCH_CONTRACT"
+     - nix-build -A stablecoin-client -o stablecoin-static
+     - nix run -f . pkgs.upx -c upx stablecoin-static/bin/stablecoin-client -o stablecoin-client
+   artifact_paths:
+     - ./stablecoin-client
 
 notify:
   - email: "tezos-alerts@serokell.io"
--- a/default.nix
+++ b/default.nix
@@ -2,19 +2,24 @@
 # SPDX-License-Identifier: MIT
 
 { sources ? import ./nix/sources.nix
+, static ? true
 , haskell-nix ? import sources."haskell.nix" { }
 , pkgs ? import sources.nixpkgs haskell-nix.nixpkgsArgs
 , weeder-hacks ? import sources.haskell-nix-weeder { inherit pkgs; }
 , ligo ? (import "${sources.ligo}/nix" { }).ligo-bin
 }:
 let
+  haskell-nix =
+    if static
+    then pkgs.pkgsCross.musl64.haskell-nix
+    else pkgs.haskell-nix;
   local-packages = [{
       name = "stablecoin";
       subdirectory = ".";
   }];
   local-packages-names = map (p: p.name) local-packages;
-  project = pkgs.haskell-nix.stackProject {
-    src = pkgs.haskell-nix.haskellLib.cleanGit {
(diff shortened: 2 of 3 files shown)


gromakovsky:
Oh, it appeared to be easier than I expected :) Will test on my machine on Friday.

gromakovsky:
The resulting executable is not statically linked.

> ./stablecoin-client: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /nix/store/8g1v3sf0xvf044sz1b4kcrg1i86z9bh7-glibc-2.30/lib/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, with debug_info, not stripped

You should use `pkgs.pkgsCross.musl64.haskell-nix` as it's done here: https://github.com/tz-wrapped/tezos-btc/blob/2a190a4b4ac414c73d8747547c11ac5a5d3274c2/tzbtc.nix#L11

dcastro:
It should be statically linked now.

gromakovsky:
Alright, works for me now. One more thing: can we `upx` the binary? It seems to work pretty well:
```
> upx stablecoin-client                                                                                                                                                              
                       Ultimate Packer for eXecutables
                          Copyright (C) 1996 - 2020
UPX git-d7ba31+ Markus Oberhumer, Laszlo Molnar & John Reiser   Jan 23rd 2020

        File size         Ratio      Format      Name
   --------------------   ------   -----------   -----------
  18412160 ->   4399612   23.90%   linux/amd64   stablecoin-client             

Packed 1 file.
```

It should be something like `nix run -f.. pkgs.upx -c upx tezos-*`.

dcastro:
Done, it's now down to 4MB.

I've also added a quick fix to my previous PR in here,

.buildkite/pipeline.yml
```diff
@@ -75,6 +75,14 @@ steps:
      # weeder needs .cabal file:
      - nix run -f.. pkgs.haskellPackages.hpack -c hpack
      - ./result
+ - label: packaging
+   if: *not_scheduled
+   commands:
+     - eval "$FETCH_CONTRACT"
+     - nix-build . -A static -o stablecoin-static
```
rvem, line 82:
```suggestion
     - nix-build -A static -o stablecoin-static
```

default.nix
```diff
@@ -53,6 +58,6 @@ in
   lib = project.stablecoin.components.library;
   test = project.stablecoin.components.tests.stablecoin-test;
   nettest = project.stablecoin.components.tests.stablecoin-nettest;
-  stablecoin-client = project.stablecoin.components.exes.stablecoin-client;
+  static = project.stablecoin.components.exes.stablecoin-client;
```
rvem, line 61: I'd still call this `stablecoin-client` since its linking way is defined by the argument that comes from the outer world

dcastro: Thanks, done!

.buildkite/pipeline.yml
dcastro: Done

PR #93, opened 2020-08-20
[#87] WIP: Add Permit to the Stablecoin contract

### LINKED ISSUE #87: Investigate and add Permit

There is this repo: https://github.com/tqtezos/lorentz-contract-permit

There was a request to add Permit to stablecoin. I haven't seen that repo before and I don't know what Permit is, so we should investigate it first.

# Acceptance criteria

1. Figure out how that repo can be used and what it does, ask questions if necessary.
2. Incorporate it into this repo.

sras:
## Related issue(s)

Resolves #87

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.


sras:
To implement this, we have to

1. Add a new storage field to store permits.
2. Add an entrypoint, which should allow users to add new permits.
3. A way to pass an permit validation procedure to the compatible entrypoints. 

When a compatible entrypoint wants to check that an authorized user has created the called parameter, it hashes the current entrypoint parameter, and calls the the lambda passed in (3), with 
the argument shown below..

```
 type sender_validation_input is (permits * blake2b_hash * address)
```

Here the `permits`, is the stored permits from storage, `blake2b_hash` is the hashed parameter,  and `address` is the address of the authorized user (like address of an admin user or a pauser).

Right now, I have started adding this to the mint entrypoint. But mint entrypoint is not compatible with this process. Because there can be many minters.

Right now 1,2 and 3 is implemented, but the implementation could be partial w.r.t what is specified in the proposal. The compatible entrypoints need to be modified, to accept the lambda sender check parameter (see the mint entrypoint as of now), and use it instead of `SENDER` to validate the sender.

dcastro:
I think this is ready for the first round of reviews.

A few things left to do: 
* Update `stablecoin-client` and add command line argument for setting the contract's default expiry (right now it's hardcoded to 1000)
* Add problem/solution details to commit messages
* Update the spec

dcastro:
Just noticed a new `GetCounter` entrypoint was added to the spec ~1 hour ago. I'll do that too.

sras (commented):
LGTM.

gromakovsky:
> Solution: modify existing entrypoints and check whether the sender is the expected user OR a permit signed by the expected user exists

Please wrap long lines in commit messages.

haskell/src/Lorentz/Contracts/Stablecoin.hs
```diff
@@ -114,7 +150,7 @@ instance Show Parameter where
   show = pretty
 
 instance ParameterHasEntrypoints Parameter where
-  type ParameterEntrypointsDerivation Parameter = EpdRecursive
+  type ParameterEntrypointsDerivation Parameter = EpdDelegate
```
gromakovsky, line 154: Why was it necessary to change `EpdRecursive` to `EpdDelegate`?

dcastro: Oh my bad, I forgot to undo that.

haskell/test/Lorentz/Contracts/Test/Permit.hs
```diff
@@ -0,0 +1,486 @@
...
+
+    specify "Permits expire after some time (set by the contract)" $ do
+      integrationalTestExpectation $ do
+        withOriginated originate defaultOriginationParams $ \stablecoinContract -> do
+          let defaultExpiry = opDefaultExpiry defaultOriginationParams
+
+          withSender testPauser $ do
+            callPermit stablecoinContract testPauserPK testPauserSK 0 Pause
+          withSender wallet1 $ do
+            rewindTime (fromIntegral defaultExpiry + 1)
```
dcastro, line 142: @gromakovsky Seems like we've found a good use case for advancing time in integration tests ^^

I'll un-deprecate this in `cleveland`, and add an `advanceTime :: Natural -> ...` function to Nettest (as part of `#53`).

gromakovsky:
I suppose the first two commits (first your commits and the only Sandeep's commit) should be squashed.

AFAIU the only remaining TODO is specification update, right?

haskell/test/Lorentz/Contracts/Test/Permit.hs
gromakovsky: > un-deprecate this

Only `rewindTime` (and rename to `advanceTime` and change its argument to be unsigned) or other time-related functions as well? I don't remember the whole story, but I think we decided that we can't reach feature parity between Nettest and Integrational, so there will be some functions only present in Integrational and it will be possible to write a test that only works with integrational. Is it right? If it's right, why did we decide to deprecate `setNow` then? Possible reasons:
1. When we decided to deprecate it, we still were hoping it would be possible to reach feature parity.
2. We decided that time manipulations other than advancing time are rarely needed, hence it's better to force users to use functions that can be implemented in both Integrational and Client nettests.

Also note that timestamps returned by `NOW` are quite descrete. AFAIU, it always returns timestamp of a block (or `f(timestamp of a block)`. So for example waiting for 10 seconds doesn't make much sense: it can be too much if blocks are created every second or can have no effect if blocks are created once a minute. So this `Natural` argument should be a number of blocks, I guess, not a time unit (like seconds). Maybe that's what you had in mind anyway.

gromakovsky:
It looks fine to me so far, but I'm going to review everything again after the specification is updated, so that I can check whether the updated spec makes sense and whether code changes comply with it.

dcastro:
> I suppose the first two commits (first your commits and the only Sandeep's commit) should be squashed.

Yes. I'll do this when squashing the "fixup" commits, when the reviews are done.

> AFAIU the only remaining TODO is specification update, right?

Yes, this is now done.

haskell/test/Lorentz/Contracts/Test/Permit.hs
dcastro: > why did we decide to deprecate setNow then

It was (2).
Back then, we decided that:

* Time-related functions were only ever used in some `nbit` tests, which could be re-written to avoid these functions. So we thought the easiest solution was to deprecate these functions, re-write the tests to avoid them and to use `Nettest` instead of `Integrational`.
* `Branchout`/`offshoot` would be added to `Nettest` in such a way that they could only be used in an emulated env.

Seeing as we've found a scenario where advancing time is _required_ (in order to test permit expiries), I propose we do for `advanceTime` what we decided to do for `branchout`/`offshoot`. (We could also close NBIT-44).

> So this Natural argument should be a number of blocks, I guess, not a time unit (like seconds).

I'm not sure what you meant here. The use case here is we want to test that a permit is no longer usable after `n` seconds have elapsed. I don't see how advancing `n` blocks would achieve this. Also, I don't see how this would be implemented in morley, since `Michelson.Runtime`\`Michelson.Interpret` have no concept of blocks.

gromakovsky: >  I propose we do for `advanceTime` what we decided to do for `branchout`/`offshoot`.

Ah, ok, I didn't realize that. So you propose to implement `advanceTime` only in Pure, not in Client. I think there is nothing preventing us from doing `advanceTime` in Client as well. We can just `threadDelay` there. The only problem is that in Client time is discrete. If you make two contract calls and want time returned by `NOW` to differ by 25 seconds (between these two calls), you can't just do `threadDelay (25 sec)` because:
1. In carthagenet blocks are created once per 30 seconds, in mainnet – once per 1 minute. So the difference can be more than 25 seconds.
2. We can say that `advanceTime x` advances time by _at least_ `x`, but we still should be careful because timestamp is taken from a block created by an arbitrary node and that node's clock can be different. I guess there is some verification of timestamps, but it's quite approximate. So if at 18:00:00 someone calls `advanceTime (1 min)` and we wait for 1 minute, we may receive block that has `18:00:59` in it (e. g. at 18:01:01) which would be incorrect. So we have to wait for a block and then maybe wait again.

So there is some complexity here, but I think it's totally possible to have `advanceTime` in both Client and Pure. And it would be nice to have it in Client because it is something that can work fine in emulated environment, but work incorrectly in real network.

There can be cases when we don't want to run tests with `advanceTime` in Client implementation. Most importantly: when it is called with a large argument, e. g. a minute or more. I think in most cases there is no need to wait for a long value, all periods can be scaled. Maybe ideally we should make it possible to say "this test will take too much time in Client, please only use Pure for it", but I guess it's not very important.

Regarding `Natural` argument – now I think it should be a time unit, and the fact that time is discrete should be handled differently (e. g. by saying that `advanceTime x` waits for `≥ x`).

dcastro:
@gromakovsky @sras I've squashed the previous commits, and pushed a few more commits with these changes:

* The `revoke` entrypoint can now only be called by the permit's issuer. Alternatively, an issuer `X` can issue a separate permit allowing others to revoke `X`'s permits. If neither of these pre-conditions hold, the entrypoint fails with `NOT_PERMIT_ISSUER`.
* When a new permit is created, the issuer's old expired permits are purged
* Added tests to ensure re-uploading the same permit hash resets its `expiry`/`created_at`.

ligo/stablecoin/spec.ligo
```diff
@@ -252,13 +302,16 @@ type ledger is big_map (address, nat)
...
-; operators                   : operators
-; transferlist_contract       : option(address)
+  ledger                  : ledger
+; token_metadata_registry : address
+; minting_allowances      : minting_allowances
+; paused                  : bool
+; roles                   : roles
+; operators               : operators
+; transferlist_contract   : option(address)
+; counter                 : counter
```
gromakovsky, line 312: I would rename it to something like `permit_counter` because just `counter` sounds a bit too generic, imo. Or maybe I would define a separate data type holding all permit-related data.

Though, since renaming a field in LIGO changes data type structure and forces us to update Haskell code, I don't insist on doing that :) But at least adding a comment would be nice.

ligo/stablecoin/actions.ligo
```diff
@@ -607,3 +708,128 @@ function set_transferlist
...
+  ) : entrypoint is block
+{ const transfer_operation : operation =
+    Tezos.transaction (store.default_expiry, 0mutez, parameter.1)
+} with
+    ( list [transfer_operation]
+    , store
+    )
+
+(*
+ * Retrieves the contract's default permit expiry and leaves the storage untouched.
```
gromakovsky, line 824: Wrong comment.

haskell/test/Lorentz/Contracts/Test/Permit.hs
```diff
@@ -370,6 +368,15 @@ permitSpec originate = do
               callPermit stablecoinContract wallet1PK wallet1SK 0 Unpause
               lCallEP stablecoinContract (Call @"Unpause") () `catchExpectedError` mgmNotPauser
 
+      specify "pauser does not consume permits" $
```
gromakovsky, line 416: There are two "pauser does not consume permits" tests, is that intended?

gromakovsky:
LGTM, besides very minor things. I'd like @sras  to re-review as well.

ligo/stablecoin/spec.ligo
dcastro: Agreed, renamed to `permit_counter` :) I also think the `revoke`/`get_counter` entrypoints could be renamed to `revoke_permit`/`get_permit_counter` in the spec.

ligo/stablecoin/actions.ligo
dcastro: Done

haskell/test/Lorentz/Contracts/Test/Permit.hs
dcastro: > is that intended

Yeah, they're under two different `describe` groups: one for the `pause` entrypoint, one for the `unpause` entrypoint.
I've renamed them to make the difference stand out.

ligo/stablecoin/spec.ligo
gromakovsky: It's something to add to TZIP-17 comments :)

gromakovsky (approved):

haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
```diff
@@ -167,14 +161,23 @@ type IsOperatorParam = View ("operator" :! OperatorParam) IsOperatorResponse
 -- solve it.
 -- TODO : https://gitlab.com/morley-framework/morley/-/issues/159
 data Parameter
-  = Transfer TransferParams
-  | Balance_of BalanceRequestParams
-  | Token_metadata TokenMetaDataParam
+  = Balance_of BalanceRequestParams
+  | Is_operator IsOperatorParam
   | Token_metadata_registry TokenMetadataRegistryParam
+  | Transfer TransferParams
```
sras, line 167: Is the ordering of constructors relevant here? If that is the case, shouldn't we leave a comment stating the same?

sras: I see that you have left the comment on the main param, so I guess it is fine.

sras:
LGTM again.

michaeljklein (approved):
LGTM

Issue #90, opened 2020-08-14
Limit the number of minters and store in a strict map


gromakovsky:
# Clarification and motivation

We store minting allowances in a `big_map`, because nothing prevents master minter from adding unbounded number of minters. However, we've agreed that in practice it's unlikely that there will be more than say 10-20 minters. So a `big_map` is overkill.

# Acceptance criteria

1. Minting allowances are stored in a strict map.
2. There can be at most 12 minters, the corresponding check is added to code. Since it's desired to keep the contract as small as possible, I think we can hardcode this constant in code. Though, if LIGO has compile-time parameters, it's better to use them.
3. The specification is updated to reflect that the number of minters is limited.
4. There is a nettest with maximal number of minters, just to be sure it doesn't break anything (12 is a small value, but maybe we'll increase it one day).

PR #95, opened 2020-08-20
[#90] Limit minters

sras:
## Description

Summary of changes.

1. Changed storage to use Map instead of big map.
2.  Added check in `configure_minter` entrypoint to see if contract already contains 12 minters, and if it does throw error "MINTER_LIMIT_REACHED" (Could probably use a better error)
3. Added tests to check if we can add minters up to the limit, and if the the contract throw error if we try to add more.
4. Added a Nettest that tries to max out the minter count.

## Related issue(s)

Resolves #98

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.


docs/specification.md
```diff
@@ -439,6 +441,9 @@ We distinguish three error cases, each has a dedicated error message:
 
 - Fails with `CONTRACT_PAUSED` if the contract is paused.
 
+- Fails with `MINTER_LIMIT_REACHED` if the current number of minters is at the limit
+  and the reqest is trying to add a new minter.
```
gromakovsky, line 445:
```suggestion
  and the request is trying to add a new minter.
```

docs/specification.md
```diff
@@ -123,6 +123,7 @@ The next group consists of the errors that are not part of the FA2 specification
 | `ADDR_NOT_MINTER`            | An attempt is made to modify minter data of an address that's not a minter                                                     |
 | `ALLOWANCE_EXCEEDED`         | Throws when trying to mint tokens more than currently allowed for an address                                                   |
 | `BAD_TRANSFERLIST`           | Given address is a not a smart contract complying with the transferlist interface                                                  |
+| `MINTER_LIMIT_REACHED`       | Cannot add new minter because the number of minters is already at the limit.
```
gromakovsky, line 126: Shouldn't this line end with `|`?

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -301,7 +307,35 @@ managementSpec originate = do
 
         mgmContractPaused err
 
+  describe "Minter limit check" $ do
+    -- Currently the contract allows to add upto 12 minters.
```
gromakovsky, line 311: I don't like that 12 (or 13 in one place) is hardcoded in so many places (including comments). Can you define a constant in just one place?

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -301,7 +307,35 @@ managementSpec originate = do
...
+    let
+      configureMinterParam :: Int -> ConfigureMinterParam
+      configureMinterParam i =
+        ( #minter .! (KeyAddress $ hashKey  $ toPublic $ detSecretKey $ encodeUtf8 @Text ("a" <> show i))
```
gromakovsky, line 315: There is `detGenKeyAddress`.

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -301,7 +307,35 @@ managementSpec originate = do
...
+
+    it "Can add minter until minter limit" $ integrationalTestExpectation $ do
+
+      withOriginated originate defaultOriginationParams $ \stablecoinContract -> do
+        withSender testMasterMinter $ do
+          mapM_ (lCallEP stablecoinContract (Call @"Configure_minter")) (configureMinterParam <$> [1..12])
+
+          lExpectStorage stablecoinContract $ \case
+            (StorageMinters minters)
+              | M.size(minters) /= 12  ->
```
gromakovsky, line 328: `M.size(minters)` looks like C++ :smiley: 

Why not `M.size minters`?

Also I think it can be made a bit shorter using `when` or `unless`, but it's a minor thing.

gromakovsky (changes requested):

gromakovsky:
Note that nettest started to fail.

haskell/test/Lorentz/Contracts/Test/Management.hs
sras: >M.size(minters) looks like C++ smiley
>Why not M.size minters?

Probably hangover from LIGO.

>Also I think it can be made a bit shorter using when or unless

Will do..

PR #96, opened 2020-08-21
[#91] Move metadata to external contract

### LINKED ISSUE #91: Move metadata into another contract

In preparation for TZIP-16 (that's currently WIP) we should extract metadata outside of our contract. So `token_metadata_registry` shouldn't return `SELF` anymore.

# Acceptance criteria

1. The `stablecoin` contract doesn't store metadata anymore. It stores an address of a contract that holds metadata instead.
2. TBD: should it be possible to update that address?
3. The repository has a simple metadata contract that only holds metadata in its storage and doesn't do anything else. I suppose we can implement it in vanilla Michelson because it's really trivial (I suppose it can always `FAILWITH` for example).
4. Tests are updated to use that metadata contract.

sras:
## Related issue(s)

Resolves #91 

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.


haskell/test/resources/metadata.tz
```diff
@@ -0,0 +1,10 @@
+{ parameter unit ;
+  storage
+    (pair (unit %dummy_field)
+          (big_map %token_metadata
+             nat
+             (pair (nat %token_id)
+                   (pair (string %symbol)
+                         (pair (string %name) (pair (nat %decimals) (map %extras string string))))))) ;
+  code { CDR ; NIL operation ; PAIR } }
+
```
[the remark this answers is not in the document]
gromakovsky: @sras I'm ok with that, but then I think we also need a CI check that `ligo compile-contract metadata.ligo` is the same as `test/resources/metadata.tz`. And we need to check that `ligo compile-contract metadata.ligo` works anyway. I believe it should be straightforward to add this check.

sras:
>Did you test that both origination scenarios work in client? 1. Manually deployed metadata contract passed from CLI. 2. No contract is passed, it's automatically originated.

Yes. This appear to work.

sras:
I have removed the `metadata.tz` from repo and have changed the pipeline. I have not merged it since I am not sure about the changed made to the nix files, and think it would be better to merge after it is reviewed.

.reuse/dep5
```diff
@@ -1,9 +1,9 @@
 Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 
-Files: .github/* nix/sources.json
+Files: .github/* nix/sources.json haskell/test/resources/metadata.tz
```
gromakovsky, line 3: This file should be reverted, I guess.

gromakovsky:
LGTM

.reuse/dep5
sras: Reverted

PR #101, opened 2020-09-03
[#97] Cleanup redundant infra and tests

### LINKED ISSUE #97: Cleanup redundant `permissions_descriptor` compatibility check infra

In recent pull requests, we removed the permission descriptor and moved token metadata contract to an external contract.
Earlier, we had an infra to check if a permission descriptor is compatible with our contract, and if not skip some tests. But since we are using an FA2 test suite catered to this contract now, we can just remove those tests. This makes the aforementioned infra redundant. so we can remove it.

# Acceptance criteria

We remove tests that require an incompatible permission descriptor and the infra that optionally run tests based on the permission descriptor is also removed.

sras:
## Description

Currently we have some tests infra that looks for a specific configuration of permissions before running the tests. But this is no longer required as we are implementing specific FA2 behavior. So we can only implement such tests and can remove the test infra for permissions configuration check.

## Related issue(s)

Resolves #97 

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/haskell/stack.yaml
+++ b/haskell/stack.yaml
@@ -42,9 +42,8 @@ extra-deps:
     https://gitlab.com/morley-framework/morley-ledgers.git
     # ^ CI cannot use ssh, so we use http clone here
   commit:
-    # TODO: update this to master
-    099082d778b8d2c25b6c9db9145f9745d0d77afe # sras/#5-sample-fa2-indigo
-  # nix-sha256: 1yy7fm5fj6m2mlrmj6c6lr5xb5l0pqq2y71ch3np7ml196iipz0i
+    47a94df200451d78d81cf0515f9328645a03bd22 # master
+  # nix-sha256: 0cx2an4kdgg9awd0y008afa6abihsy6rqldq2xxmyb2m20vdzgms
   subdirs:
     - code/morley-ledgers
     - code/morley-ledgers-test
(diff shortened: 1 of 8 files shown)


haskell/test-common/Lorentz/Contracts/Test/Common.hs
```diff
@@ -148,19 +148,16 @@ constructSingleTransfer from to amount
...
 withOriginated
   :: forall param. OriginationFn param
   -> OriginationParams
   -> (TAddress param -> IntegrationalScenario)
   -> IntegrationalScenario
-withOriginated fn op tests = do
-  (fn op) >>= \case
-    Nothing -> pass
-    Just contract -> tests contract
+withOriginated fn op tests = (fn op) >>= tests
```
gromakovsky, line 158:
```suggestion
withOriginated fn op tests = fn op >>= tests
```

gromakovsky (approved):

haskell/stack.yaml
```diff
@@ -43,8 +43,8 @@ extra-deps:
     # ^ CI cannot use ssh, so we use http clone here
   commit:
     # TODO: update this to master
```
gromakovsky, line 45: This TODO looks obsolete now.

sras: Removed.

PR #106, opened 2020-09-14
Nightly delphinet runs

rvem:
## Description
Problem: We want to test our contracts and library on new testnet.

Solution: Add scheduled nettest run on delphinet.

## Related issue(s)

https://issues.serokell.io/issue/TM-455

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -4,6 +4,7 @@
 env:
   FETCH_CONTRACT:
     "mkdir -p haskell/test/resources/ && buildkite-agent artifact download stablecoin.tz haskell/test/resources/ && buildkite-agent artifact download metadata.tz haskell/test/resources/ --step \"LIGO-contract\""
+  TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER: "Y"
 
 steps:
  - label: hlint
@@ -44,7 +45,6 @@ steps:
  - label: nettest-local-chain
    if: *not_scheduled
    env:
-     TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER: "Y"
      NETTEST_NODE_ADDR: "localhost"
      NETTEST_NODE_PORT: "8734"
      MONEYBAG: "unencrypted:edsk3GjD83F7oj2LrnRGYQer99Fj69U2QLyjWGiJ4UoBZNQwS38J4v"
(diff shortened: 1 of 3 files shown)


nix/sources.json
```diff
@@ -66,15 +66,15 @@
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
     "tezos-packaging": {
-        "branch": "v7.0-1",
+        "branch": "master",
```
gromakovsky, line 69: I guess we can use a tag here again.

gromakovsky: Though I am not sure it's better than `master` :shrug:

.buildkite/pipeline.yml
```diff
@@ -68,6 +68,18 @@ steps:
...
    timeout_in_minutes: 120
+ - label: nettest-scheduled-delphinet
+   if: build.source == "schedule"
+   env:
+     TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER: "Y"
+     NETTEST_NODE_ADDR: "delphi.testnet.tezos.serokell.team"
+     NETTEST_NODE_PORT: "8732"
+     # Note that testnet moneybag can run out of tz. If this happened, someone should transfer it some
+     # more tz, its address: tz1ZLfAd3LVQUdj1vEX8YivD1B3sE23F29Fd
+     MONEYBAG: "unencrypted:edsk4Aq1Q3fc1ehMKTL55wcJxdz52r1iJggPHFkSFq57toaNNFVxKV"
```
gromakovsky, line 79: It may be a bit simpler to use the same address in both *nets (e. g. we won't have to write similar comments twice), but I don't insist on doing that.

gromakovsky: I mean you can transfer from `tz1ZLfAd3LVQUdj1vEX8YivD1B3sE23F29Fd` to `tz1ij8gUYbMRUXa4xX3mNvKguhaWG9GGbURn`.

.buildkite/pipeline.yml
```diff
@@ -68,6 +68,18 @@ steps:
    commands: *run-nettest
    retry: *retry-nettest
    timeout_in_minutes: 120
+ - label: nettest-scheduled-delphinet
+   if: build.source == "schedule"
+   env:
+     TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER: "Y"
```
gromakovsky, line 74: Can we extract it to global variables btw?

gromakovsky (approved):
LGTM, I have only some minor suggestions.

nix/sources.json
rvem: Yes, we can use tag now, since 7.4 tag was released

PR #108, opened 2020-09-17
[#107] Write Nettest tests for the new permit features

dcastro:
## Description

Added integration tests for the new permit features that run on a real network.

## Related issue(s)

Resolves #107

#### Related changes (conditional)

- Tests
  - [x] If I added new functionality, I added tests covering it.
  - [x] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
  - [x] I checked whether I should update the docs and did so if necessary:
    - [README](../tree/master/README.md)
    - Haddock
  - [x] I updated [changelog](../tree/master/ChangeLog.md) unless I am sure my changes are
        not essential.

#### Stylistic guide (mandatory)

- [x] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).
- [x] My code complies with the [style guide](../tree/master/docs/code-style.md).

PR #112, opened 2020-09-24
[#102] Use morley capabilities to define custom ADT layout

### LINKED ISSUE #102: Use morley capabilities to define custom ADT layout

We have types like
```
type Storage =
  ((((Counter, DefaultExpiry),
     (Ledger, MintingAllowances)),
    ((Operators, IsPaused),
     (Permits, Roles))),
   (TokenMetadataRegistryAddress, TransferlistContract))
```

That's insane and is needed only to ensure layout compatible with LIGO code. Fortunately now there is `customGeneric` in Lorentz, so we can turn it into a normal ADT and use `customGeneric` to enforce a specific layout.

# Acceptance criteria

We don't have data types defined as a lot of nested tuples. We define them as ADTs and use `customGeneric` to specify their layout.

dcastro:
## Description

Problem: The Michelson representation (that is, the balancing of `pair`
and `or`s) of our Haskell data types must match the Michelson
representation of our Ligo data types.

In the past, we used Haskell nested tuples in order to have complete
control of the layout, but this leads to messy, hard to maintain code.

Solution: Now that morley introduced `customGeneric` to control an ADT's
layout, we can replace our nested tuples with Haskell ADTs and still
have control of its layout.

## Related issue(s)

Resolves #102 

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
+++ b/haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
@@ -6,9 +6,9 @@
 -- https://gitlab.com/tzip/tzip/-/blob/131b46dd89675bf030489ded9b0b3f5834b70eb6/proposals/tzip-12/tzip-12.md
 
 module Lorentz.Contracts.Spec.FA2Interface
-  ( BalanceRequestItem
+  ( BalanceRequestItem(..)
   , BalanceRequestParams
-  , BalanceResponseItem
+  , BalanceResponseItem(..)
   , CustomPermissionPolicy
   , FA2OwnerHook (..)
   , IsOperatorParam
@@ -69,12 +69,22 @@ type TransferParams = [TransferParam]
 -- ---------
 -- Queries balance of one or more addresses. Callback contract should accept a list of BalanceResponse
 -- Duplicates in the request should not be de-duplicated or reordered in the response.
-type BalanceRequestItem = ("owner" :! Address, "token_id" :! TokenId)
+data BalanceRequestItem = BalanceRequestItem
+  { briOwner :: Address
+  , briTokenId :: TokenId
+  }
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (IsoValue, HasAnnotation)
 
 instance Buildable BalanceRequestItem where
   build = genericF
 
-type BalanceResponseItem = ("request" :! BalanceRequestItem, "balance" :! Natural)
+data BalanceResponseItem = BalanceResponseItem
+  { briRequest :: BalanceRequestItem
+  , briBalance :: Natural
+  }
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (IsoValue, HasAnnotation)
 
 type BalanceRequestParams = View ("requests" :! [BalanceRequestItem]) [BalanceResponseItem]
 
--- a/haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
+++ b/haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
@@ -19,7 +19,7 @@ module Lorentz.Contracts.Spec.FA2Interface
   , Parameter (..)
   , FA2ParameterC
   , TokenId
-  , TokenMetadata
+  , TokenMetadata(..)
   , TransferDestination
   , TransferDescriptor
   , TransferDescriptorParam
@@ -29,7 +29,7 @@ module Lorentz.Contracts.Spec.FA2Interface
   , UpdateOperatorsParam
   ) where
 
-import Fmt (Buildable(..), genericF, (+|), (|+))
+import Fmt (Buildable(..), genericF)
 import Test.Tasty.QuickCheck (Arbitrary(..), elements)
 
 import Lorentz
@@ -69,26 +69,32 @@ instance Buildable BalanceResponseItem where
   build = genericF
 
 -- Token MetaData query
+data TokenMetadata = TokenMetadata
+ { tmTokenId :: TokenId
+ , tmSymbol :: MText
+ , tmName :: MText
+ , tmDecimals :: Natural
+ , tmExtras :: Map MText MText
+ }
+ deriving stock Show
 
-type TokenMetadata =
-  ( "token_id" :! TokenId
-  , "mdr" :! ("symbol" :! MText
-  , "mdr2" :! ("name" :! MText
-  , "mdr3" :! ("decimals" :! Natural
-  , "extras" :! Map MText MText)))
+instance Buildable TokenMetadata where
+  build = genericF
+
+deriving anyclass instance IsoValue TokenMetadata
+deriving anyclass instance HasAnnotation TokenMetadata
+$(customGeneric "TokenMetadata" $ withDepths
+    [ cstr @0
+      [ fld @1
+      , fld @2
+      , fld @3
+      , fld @4
+      , fld @4
+      ]
+    ]
   )
 
 type TokenMetadataRegistryParam = ContractRef Address
-
-instance Buildable TokenMetadata where
-  build
-    ( arg #token_id -> token_id
-    , arg #mdr -> (arg #symbol -> symbol
-    , arg #mdr2 -> (arg #name -> name
-    , arg #mdr3 -> (arg #decimals -> decimals
-    , _)))) =
-    "token_id:" +| token_id |+ ", symbol:" +| symbol |+ ", name:" +|
-    name |+ ", decimals:" +| decimals |+ "."
 
 data OperatorTransferPolicy
   = OwnerTransfer ("owner_transfer" :! ())
--- a/haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
+++ b/haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
@@ -9,7 +9,7 @@ module Lorentz.Contracts.Spec.FA2Interface
   ( BalanceRequestItem(..)
   , BalanceRequestParams
   , BalanceResponseItem(..)
-  , CustomPermissionPolicy
+  , CustomPermissionPolicy(..)
   , FA2OwnerHook (..)
   , IsOperatorParam
   , IsOperatorResponse(..)
@@ -155,8 +155,12 @@ instance Arbitrary OwnerTransferMode where
 instance TypeHasDoc OwnerTransferMode where
   typeDocMdDescription = "Describes if owener hooks are required in sender/receiver addresses"
 
-type CustomPermissionPolicy
-  = ("tag" :! MText, "config_api" :! Maybe Address)
+data CustomPermissionPolicy = CustomPermissionPolicy
+  { cppTag :: MText
+  , cppConfigApi :: Maybe Address
+  }
+  deriving stock (Generic, Show, Eq)
+  deriving anyclass (IsoValue, HasAnnotation)
 
 data OperatorParam = OperatorParam
   { opOwner :: Address
--- a/haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
+++ b/haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
@@ -21,8 +21,9 @@ module Lorentz.Contracts.Spec.FA2Interface
   , TokenId
   , TokenMetadata(..)
   , TransferDestination(..)
-  , TransferDescriptor
-  , TransferDescriptorParam
+  , TransferDestinationDescriptor(..)
+  , TransferDescriptor(..)
+  , TransferDescriptorParam(..)
   , TransferParams
   , TransferParam(..)
   , UpdateOperator (..)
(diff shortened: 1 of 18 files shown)


haskell/test/Lorentz/Contracts/Test/FA2.hs
```diff
@@ -246,11 +246,10 @@ fa2Spec fa2Originate = do
 
         withSender commonOperator $ lCallEP stablecoinContract (Call @"Transfer") transfer1
 
-        lExpectStorage stablecoinContract $ \case
-          (StorageLedger ledger)
-            | M.lookup wallet1 ledger == Nothing -> Right ()
-            | otherwise ->
-                Left $ CustomTestError "Zero balance account was not removed"
+        lExpectStorage stablecoinContract $ \storage ->
+          if M.lookup wallet1 (unBigMap $ sLedger storage) == Nothing
```
gromakovsky, line 248: I would use `unless` or `when` here (and in some other places), but that's up to you.

UPD: looks like there are many such places, so probably it's not worth changing them now :)

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -99,20 +99,19 @@ managementSpec originate = do
...
       withOriginated originate defaultOriginationParams $ \stablecoinContract -> do
         withSender testPauser $ lCallEP stablecoinContract (Call @"Pause") ()
-        lExpectStorage stablecoinContract $ \case
-          StoragePaused isPaused
-            | not isPaused ->
-                Left $ CustomTestError "Contract is not paused as was expected"
-            | otherwise -> Right ()
+        lExpectStorage stablecoinContract $ \storage ->
+          if sIsPaused storage
+            then Right()
```
gromakovsky, line 113:
```suggestion
            then Right ()
```

here and in at least one more place.

haskell/src/Lorentz/Contracts/Stablecoin.hs
```diff
@@ -62,26 +61,34 @@ metadataRegistryContractPath = "./test/resources/metadata.tz"
...
-  ( arg #minter -> addr
-  , (arg #current_minting_allowance -> cma, arg #new_minting_allowance -> nma))
-{-# COMPLETE ConfigureMinterParams #-}
+deriving anyclass instance IsoValue ConfigureMinterParam
+deriving anyclass instance HasAnnotation ConfigureMinterParam
+$(customGeneric "ConfigureMinterParam" $ withDepths
+    [ cstr @0
+      [ fld @1
+      , fld @2
+      , fld @2
```
gromakovsky, line 77: Apparently it's the same structure as the one that would be produced by vanilla `Generic` deriving. Actually I don't know what is better: explicitly write that we need this structure or prefer a shorter version that implicitly generates this structure. Both are somewhat fragile anyway :yao:

So I am suggesting this change, but I am not sure it's better than the current version, so feel free to ignore it.

haskell/src/Lorentz/Contracts/Stablecoin.hs
```diff
@@ -62,26 +61,34 @@ metadataRegistryContractPath = "./test/resources/metadata.tz"
...
+instance Buildable ConfigureMinterParam where
+  build = genericF
 
-pattern ConfigureMinterParams :: Address -> Maybe Natural -> Natural -> ConfigureMinterParam
-pattern ConfigureMinterParams addr cma nma <-
-  ( arg #minter -> addr
-  , (arg #current_minting_allowance -> cma, arg #new_minting_allowance -> nma))
-{-# COMPLETE ConfigureMinterParams #-}
+deriving anyclass instance IsoValue ConfigureMinterParam
+deriving anyclass instance HasAnnotation ConfigureMinterParam
```
gromakovsky, line 72: Why does it have to be standalone btw?

haskell/src/Lorentz/Contracts/Stablecoin.hs
dcastro: Yeah, I struggled with the same dilemma :sweat: 

I went for explicit-style, but I agree, both approaches are equally fragile. So now I'm starting to think we should go for the shorter version. I'll make this change.

Regarding the maintenance of these layouts: in #108, I changed [`parseStablecoinContract`](https://github.com/tqtezos/stablecoin/pull/108/files#diff-102dc12d0013cd28fb3bad312ffd8918L26) to return a typed contract instead of an untyped contract. `parseStablecoinContract` is used extensively in the test suites - so if a Generic layout goes out of sync, the tests are sure to catch this.

We can even add a compile-time check for this. With about ~5 lines of TH, we could typecheck the contract at compile-time and ensure the layouts match. What do you think? Or is this too much?

haskell/src/Lorentz/Contracts/Stablecoin.hs
dcastro: Otherwise, I get these errors:

```
• No instance for (WellTyped
                     (GValueType (Rep ConfigureMinterParam)))
    arising from the 'deriving' clause of a data type declaration
  Possible fix:
    use a standalone 'deriving instance' declaration,
      so you can specify the instance context yourself
• When deriving the instance for (IsoValue ConfigureMinterParam)
```
```
• No instance for (GHasAnnotation (Rep ConfigureMinterParam))
    arising from the 'deriving' clause of a data type declaration
  Possible fix:
    use a standalone 'deriving instance' declaration,
      so you can specify the instance context yourself
• When deriving the instance for (HasAnnotation
                                    ConfigureMinterParam)
```

It seems it's not able to find the `Generic` instance. Maybe it's because it's being generated using TH?

haskell/test/Lorentz/Contracts/Test/Management.hs
dcastro: Done

haskell/test/Lorentz/Contracts/Test/FA2.hs
dcastro: It didn't take long to change, done :)

haskell/src/Lorentz/Contracts/Stablecoin.hs
dcastro: > So now I'm starting to think we should go for the shorter version. I'll make this change.

Done

dcastro: Correction: 4 lines of TH :smile:  

```haskell
$(case readContract @(ToT Parameter) @(ToT Storage) stablecoinPath $(embedStringFile stablecoinPath) of
      Left e -> fail (pretty e)
      Right _ -> pure []
 )
```

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -758,7 +758,7 @@ managementSpec originate = do
         withOriginated originate originationParams $ \stablecoinContract -> do
           let
             transfers =
-              [(#from_ .! wallet1, #txs .! [(#to_ .! wallet2, (#token_id .! 0, #amount .! 10))])]
+              [FA2.TransferParam { tpFrom = wallet1, tpTxs = [FA2.TransferDestination { tdTo = wallet2, tdTokenId = 0, tdAmount = 10 }]}]
```
gromakovsky, line 750: Too long line.

haskell/test/SMT.hs
```diff
@@ -710,25 +709,22 @@ applyTransfer ccSender storage tis = do
...
-          then let
-            in foldl' singleTranferTx (Right storage) toItems
-          else Left FA2_NOT_OPERATOR
+applySingleTransfer ccSender estorage (FA2.TransferParam from txs) =
+  case estorage of
+    Left err -> Left err
+    Right storage@(SimpleStorage {..}) ->
+      if ccSender == from || isOperatorOf storage ccSender from
+        then let
+          in foldl' singleTranferTx (Right storage) txs
```
gromakovsky, line 718: You didn't change it in this PR, but I am looking at it now and don't understand this construct. AFAIU there is empty `let in` block which is redundant. If it is, can you remove it please while we are at it?

haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
```diff
@@ -240,23 +241,58 @@ type FA2ParameterC param =
...
+$(customGeneric "TransferDescriptorParam" $ withDepths
+    [ cstr @0
+      [ fld @1
+      , fld @2
+      , fld @2
+      ]
+    ]
+  )
 
 data FA2OwnerHook
```
gromakovsky, line 271: Can you please check whether hook-related types are used at all? I suspect they may be not used anymore because we are using default permission descriptor now which has "no hook" policy.

haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
```diff
@@ -165,8 +165,12 @@ instance Arbitrary OwnerTransferMode where
 instance TypeHasDoc OwnerTransferMode where
   typeDocMdDescription = "Describes if owener hooks are required in sender/receiver addresses"
 
-type CustomPermissionPolicy
-  = ("tag" :! MText, "config_api" :! Maybe Address)
+data CustomPermissionPolicy = CustomPermissionPolicy
```
gromakovsky, line 158: Hm, apparently this type is unused. I think I'll create a new issue about finding unused types are removing them.

haskell/src/Lorentz/Contracts/Stablecoin.hs
gromakovsky: Okay, if there is a compile error with normal deriving, I don't mind using standalone deriving, not a problem.

> Maybe it's because it's being generated using TH?

Yes, probably.

haskell/src/Lorentz/Contracts/Stablecoin.hs
gromakovsky: > Or is this too much?

I think 4 lines of TH is not too much, I'm ok with doing that :)

haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
dcastro: `FA2OwnerHook` seems to only be used in a few tests in `Lorentz.Contracts.Test.FA2`

There's this comment next to those tests:

```
  ---- Owner hook test
  ----
  ---- Tests that validate that senders owner hook is not-called on transfer
  ---- using default origination options where both sender/receiver hooks
  ---- are set to be not called no matter what.
```

haskell/test/SMT.hs
dcastro: Done

haskell/test/Lorentz/Contracts/Test/Management.hs
dcastro: Done

haskell/src/Lorentz/Contracts/Stablecoin.hs
dcastro: Done

haskell/src/Lorentz/Contracts/Stablecoin.hs
```diff
@@ -347,3 +346,46 @@ deriving anyclass instance IsoValue (MetadataRegistryStorageView)
...
+-- This can usually be fixed by writing a custom instance of Generic using `customGeneric`,
+-- and making sure the data type's layout matches the layout in stablecoin.tz.
+-- See examples above.
+$(case readContract @(ToT Parameter) @(ToT Storage) stablecoinPath $(embedStringFile stablecoinPath) of
+    Left e -> fail (pretty e)
+    Right _ -> pure []
+ )
+
+-- | Parse the stablecoin contract.
+parseStablecoinContract :: MonadThrow m => m (T.Contract (ToT Parameter) (ToT Storage))
```
gromakovsky, line 378: If we have that splice above, we know in compile time that the contract is valid (otherwise compilation fails). Can we use this knowledge here to avoid `MonadThrow` and change it to just `T.Contract (ToT Parameter) (ToT Storage)`? I am bad at TH, but I suspect it should be straightforward.

haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
gromakovsky: Hm, okay, perhaps that test makes some sense. So I've created an issue that covers unused types as well as types present in `morley-ledgers`: https://github.com/tqtezos/stablecoin/issues/115

haskell/src/Lorentz/Contracts/Spec/FA2Interface.hs
gromakovsky: https://github.com/tqtezos/stablecoin/issues/115

haskell/src/Lorentz/Contracts/Stablecoin.hs
dcastro: I don't _think_ we can, or at least not in that way.

The TH splice would have to compile the contract (already done), and then somehow emit a valid [haskell expression](https://hackage.haskell.org/package/template-haskell-2.16.0.0/docs/Language-Haskell-TH.html#t:Exp) of type `T.Contract (ToT Parameter) (ToT Storage)`, as if it had been handwritten (not easily done).

In other words, what would the code generated by TH look like?

```
stablecoinContract :: T.Contract (ToT Parameter) (ToT Storage)
stablecoinContract = ???
```

Theoretically, the TH splice could pattern match on the parsed `T.Contract (ToT Parameter) (ToT Storage)`, and build an `Exp` AST representing calls to `Michelson.Typed.Instr` and `Michelson.Typed.Value`. It is doable, but it's a lot of work for little gain.

However, because the TH splice proves the contract is well-typed, we can change `either throwM` to `either error`, and get rid of `MonadThrow`, and be sure that `error` will never be called.

gromakovsky: Okay

> However, because the TH splice proves the contract is well-typed, we can change `either throwM` to `either error`, and get rid of MonadThrow, and be sure that error will never be called.

I think it's a good idea (as long as there is a comment explaining why the error can't happen here). It matches the idea of using `error` for errors that can happen only in case of a programmer mistake.

dcastro: Done, looks a lot better now!

gromakovsky: :cool:

gromakovsky (approved):

PR #119, opened 2020-09-29
Run nettest jobs on a separate agent to not stall CI for regular jobs

zhenyavinogradov:
## Description

Problem: scheduled nettest job takes a long time to finish, making other
CI jobs wait for it

Solution: use a special runner for the nettest job

Needs serokell/serokell-profiles#540 to be deployed first

## Related issue(s)

https://issues.serokell.io/issue/OPS-1066

#### Related changes (conditional)

- Tests
  - [ ] If I added new functionality, I added tests covering it.
  - [ ] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
  - [ ] I checked whether I should update the docs and did so if necessary:
    - [README](../tree/master/README.md)
    - Haddock
  - [ ] I updated [changelog](../tree/master/ChangeLog.md) unless I am sure my changes are
        not essential.

#### Stylistic guide (mandatory)

- [ ] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).
- [ ] My code complies with the [style guide](../tree/master/docs/code-style.md).

PR #126, opened 2020-10-12
CI: add 'wait' step after uploading generated contract

zhenyavinogradov:
## Description

Problem: 'LIGO-contract' step uploads an artifact used by the next
steps, so the next steps must not start until it has finished. It wasn't
a problem when we were using a single non-concurrent agent, but when we
added an agent for scheduled pipelines in #119, the job would fail because it
runs concurrently.

Solution: add 'wait' step to avoid running steps which use the contract
until the step which generates the contract has finished
## Related issue(s)

https://issues.serokell.io/issue/OPS-1066

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -38,6 +38,9 @@ steps:
      - stablecoin.fa1.2.tz
      - metadata.tz
 
+ # wait for the contract step to complete, so the next steps can download generated contract
+ - wait
+
  - label: build library
    if: *not_scheduled
    commands:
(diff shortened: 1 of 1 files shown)


rvem (approved):

rvem:
Hmm, looks like `nettest-scheduled-delphinet` doesn't have appropriate queue tag:(

zhenyavinogradov:
Now it does :007:

rvem (approved):

PR #128, opened 2020-10-14
[#127] Migrate SMT module to use Hedgehog

### LINKED ISSUE #127: Migrate SMT to hedgehog

Our state machine tests currently use `QuickCheck` as the underlying library. However, QuickCheck support is being [removed](https://gitlab.com/morley-framework/morley/-/merge_requests/610) from `morley` and `cleveland` in favor of `hedgehog`.

# Acceptance criteria

State machine tests use `hedgehog` instead of `QuickCheck`.

gromakovsky:
## Description

Problem: our state machine tests currently use `QuickCheck` as the
underlying library. However, QuickCheck support is being
[removed](https://gitlab.com/morley-framework/morley/-/merge_requests/610)
from `morley` and `cleveland` in favor of `hedgehog`.
In particular, `integrationalTestProperty` is being removed, along with
`Arbitrary` instances.
We should stop using these functions to be able to migrate to newer
versions of those libraries.

Solution: mechanically replace QuickCheck usages to their hedgehog
counterparts.

## Related issue(s)

Resolves #127

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/haskell/test/SMT.hs
+++ b/haskell/test/SMT.hs
@@ -749,5 +749,8 @@ applyFA2Parameter ContractCall {..} cs = do
     FA2.Update_operators ops -> foldl' (applyUpdateOperator ccSender) (Right cs) ops
     _ -> error "Unexpected param"
 
+-- | Generate a list of items using given generator of a single item.
+-- Takes a size parameter. Unlike QuickCheck's @vectorOf@, this function
+-- can generate a smaller list (always non-empty).
 vectorOf :: MonadGen m => Int -> m x -> m [x]
-vectorOf n = Gen.list (Range.singleton n)
+vectorOf n = Gen.list (Range.linear 1 n)
--- a/haskell/test/SMT.hs
+++ b/haskell/test/SMT.hs
@@ -1,7 +1,6 @@
 -- SPDX-FileCopyrightText: 2020 TQ Tezos
 -- SPDX-License-Identifier: MIT
 {-# OPTIONS_GHC -Wno-orphans #-}
-{-# OPTIONS_GHC -Wno-deprecations #-}
 {-# LANGUAGE PackageImports #-}
 
 module SMT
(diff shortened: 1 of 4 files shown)


haskell/test/SMT.hs
```diff
@@ -748,3 +752,6 @@ applyFA2Parameter ContractCall {..} cs = do
     FA2.Transfer tis -> applyTransfer ccSender cs tis
     FA2.Update_operators ops -> foldl' (applyUpdateOperator ccSender) (Right cs) ops
     _ -> error "Unexpected param"
+
+vectorOf :: MonadGen m => Int -> m x -> m [x]
+vectorOf n = Gen.list (Range.singleton n)
```
dcastro, line 757: Would it be better to use `Range.linear` or  `Range.constant`, so that we get shrinked values when a test fails?

haskell/test/SMT.hs
```diff
@@ -11,25 +10,29 @@ module SMT
...
-  vectorOf)
+import Hedgehog
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
 import qualified Text.Show
 
-import Tezos.Core (unsafeMkMutez)
+-- import Test.Tasty.QuickCheck
+--   (Arbitrary, Gen, Property, arbitrary, arbitraryBoundedEnum, choose, elements, shuffle, sublistOf,
+--   vectorOf)
```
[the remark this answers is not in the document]
gromakovsky: Right, forgot to do that.

haskell/test/SMT.hs
gromakovsky: I am not sure, but AFAIU this behavior is equivalent to what we had in QuickCheck version, because AFAIU `vectorOf` doesn't shrink the size and doesn't depend on the size parameter. However, it seems safe to shrink here, so I guess I'll change it to `linear` in a separate commit.

dcastro (approved):

PR #130, opened 2020-10-26
Restore retries for scheduled runs

gromakovsky:
## Description

Problem: we used to retry failed scheduled nettest runs on CI
beucase they are fragile, but we stopped doing that at some point
because they could stall CI completely.
Now our scheduled nettests are launched on a separate agent, so
it shouldn't be a problem to restart them once in case of failure.

Solution: add `retry` to two jobs running scheduled nettests.

## Related issue(s)

None

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.


rvem:
Let's also increase timeout once again, looks like recent failures mostly caused by it

rvem (approved):

Issue #124, opened 2020-10-08
Update TZIP-17 implementation


dcastro:
# Clarification and motivation

In #93, we implemented TZIP-17 in Stablecoin. 
Since then, we submitted some feedback [here](https://gitlab.com/tzip/tzip/-/merge_requests/83), and the TZIP has been revised.

For example:
* the TZIP now states that when a permit is issued twice, the contract should fail with `"DUP_PERMIT"`.
* the `revoke` entrypoint has been removed
* TZIP-16 compliance is now required
* `setExpiry`'s parameter type has changed
* a user can issue a permit for the `setExpiry` entrypoint
* there should be an upper bound limit on how big an expiry can be (e.g. 10^10 ~ 3000 years)

We should update our implementation to reflect these changes.

# Acceptance criteria

Stablecoin is TZIP-17-compliant.

PR #135, opened 2020-11-25
[#124] Implemented TZIP-16 metadata

### LINKED ISSUE #109: Delete ChainID workaround in Nettest tests

Because Nettest does not currently support retrieving the ChainID, we had to use a [workaround](https://github.com/tqtezos/stablecoin/blob/914392f6c3cdfe4cefeca784eb72c3baa6555d79/haskell/nettest/Permit.hs#L143-L161) in #108 to do this.

Once [morley/#356](https://gitlab.com/morley-framework/morley/-/issues/356) is implemented, we can delete this workaround

# Acceptance criteria

The workaround we use to retrieve the ChainId in the Permit nettests is deleted, and we're using the new function in Cleveland instead.

dcastro:
## Description

Problem: There have been recent changes to TZIP-17, and the spec now
mandates contracts also implement TZIP-16.

Solution: Implement TZIP-16. In particular, we:

* Added a metadata big_map
* Moved the "GetCounter" and "GetDefaultExpiry" views off the contract's
  code. They now live inside the metadata JSON, which, at the moment, is
  stored in the contract's storage (but this could change).

## Related issue(s)

Resolves part of #124

Resolves #109

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.


dcastro:
A few notes about my design decisions:

1. The TZIP-16 metadata can be stored 1) in the contract's storage, 2) in a separate contract's storage, or 3) off-chain in a http/https/ifps location. I chose to store it in the contract's storage as it's the simplest approach, but I can change this if needed.
2. The spec also mentions an optional `assertMetadata` entrypoint(s). I did not implement these, but I can if needed as well.

-----

The current `TZIP16Interface` implementation is not fully generalized, but it is sufficient to fit the needs of stablecoin. In particular:

1. it should be moved to a reusable package
1. it does not preserve the annotations of the view's parameter and return types (but the "GetCounter" and "GetDefaultExpiry" views don't have annotations anyway)
1. I did not add support for "Rest API Views", only "Michelson Storage Views"
1. I did not add support for dynamic errors, only static errors
1. the json deserializers will probably fail if optional fields like `authors` are not present
1. the interface should be expanded to make it easier to retrieve the contract's metadata
1. from a dev experience point-of-view, it's possible to make the mistake of decoding the metadata json to the wrong type, or calling a view with the wrong parameter type/return type. we should look for ways to statically ensure this does not happen.

------

Problems I found and I'd like to discuss:

1. With the introduction of TZIP-16, views must now be run off-chain, and you must pass them the contract's storage.
This means
    1. We cannot run stablecoin's views in nettest, because we cannot retrieve storages that contain big_maps.
    2. Even when we do, the storage will contain big_map IDs instead of the big_map contents, and our interpreter can't handle that. I suppose we'd need an interpreter that is capable of fetching big map values from an actual chain.

Running views off-chain seems like a big burden on users, I suspect the mid/long-term goal is to add support to `tezos-client` so it does this automatically, but I'm not sure what the actual plans are. Or does this exist already and I'm not aware of it? :O

gromakovsky:
Regarding part 1 (storing metadata outside of the contract and optional `assertMetadata`) let's do the following: 
1. Stick to the simplest approach for now because we need full TZIP-17 compliance ASAP.
2. I think metadata should be moved out of the contract (e. g. we were previously asked to do that for FA2 metadata), but I am not sure where exactly (another contract or off-chain). AFAIU, it doesn't affect the stablecoin contract anyway, it only matters for `stablecoin-client`, right? We can update `stablecoin-client` to support 2 on-chain ways of storing metadata (in a separate issue later). Off-chain way sounds harder to setup, so I wouldn't worry about it unless explicitly asked. I mean, making sure that something is accessible in a http/https/ifps location sounds out of scope for `stablecoin-client`.

gromakovsky:
> The current TZIP16Interface implementation is not fully generalized, but it is sufficient to fit the needs of stablecoin. In particular:

I think the right place for this new reusable package is in https://gitlab.com/morley-framework, e. g. that's where we have `morley-ledgers`, `morley-upgradeable`. I'll create such a repo and will make an issue to actually implement such a package (copying your notes there).

gromakovsky:
> We cannot run stablecoin's views in nettest, because we cannot retrieve storages that contain big_maps.

That's sad, but AFAIU we can at least relatively easily add such support to emulator nettest and test off-chain views using only emulator for now.

> Running views off-chain seems like a big burden on users, I suspect the mid/long-term goal is to add support to tezos-client so it does this automatically, but I'm not sure what the actual plans are. Or does this exist already and I'm not aware of it? :O

I see that TZIP-16 refers to https://gitlab.com/smondet/tezos/-/merge_requests/7 which is said to support
> “calling” off-chain-views

There is an example:
```
 $ tezos-client --endpoint https://carthagenet.SmartPy.io:443 \
         call-off-chain-view multiply-the-nat \
         from KT1XRT495WncnqNmqKn4tkuRiDJzEiR4N2C9 \
         --arg 3
39
```

I don't know if it can handle `big_map`s fully, but I think it should eventually (maybe it's not implemented yet).

haskell/nettest/Permit.hs
```diff
@@ -129,32 +123,12 @@ permitScenario = uncapsNettest $ do
...
--- | A contract that stores the current ChainId in its storage.
---
--- NOTE: `compileLorentzContract` requires the contract's parameter to have an instance for
--- `HasAnnotation`, and ChainId doesn't.
--- So, instead, we store it as a ByteString and pack/unpack as needed.
-chainIdContract :: Contract () ByteString
-chainIdContract = defaultContract $
-  L.drop # L.chainId # L.pack # L.nil # L.pair
+      alias <- getAlias (AddressResolved addr)
+      pure (alias, addr)
```
gromakovsky, line 139: Oh, apparently it resolves https://github.com/tqtezos/stablecoin/issues/109

dcastro: Yup, I added #109 to this PR's description now :)

haskell/src/Stablecoin/Util.hs
```diff
@@ -0,0 +1,51 @@
...
+import qualified Michelson.Typed as T
+import Util.Default (Default(def))
+import Util.Typeable ((:~:)(Refl), eqT)
+
+aesonOptions :: Options
+aesonOptions = (aesonPrefix camelCase) { omitNothingFields = True}
+
+-- | Similar to 'Michelson.Test.Import.readContract',
+-- but instead of showing just the expected type @cp@ when an error occurs,
+-- it shows both the expected @cp@ and unexpected @cp'@ types.
```
gromakovsky, line 26: I see the following code snippet which AFAIU is supposed to be print both types:
```
instance Buildable ImportContractError where
  build =
    \case
      ICEUnexpectedParamType actual expected ->
        "Unexpected parameter type: " +| actual |+
        ", expected: " +| expected |+ ""
```

Does it work incorrectly there? Or am I reading it wrong?

gromakovsky:
Btw, `xrefcheck` reports some errors now, apparently they are caused by changed in this PR.

haskell/test/Lorentz/Contracts/Test/Permit.hs
```diff
@@ -82,6 +89,65 @@ assertPermitCount contractAddr expectedCount =
...
+
+    when (actualViewVal /= expectedViewVal) $
+      Left $ CustomTestError $ unlines
+        [ "Expected: "
+        , show expectedViewVal
+        , "Got:"
+        , show actualViewVal
+        ]
+  where
+    onNothing :: Text -> Maybe a -> Either TestError a
```
gromakovsky, line 147: There is [`maybeToRight`](https://hackage.haskell.org/package/universum-1.7.1/docs/Universum-Monad-Either.html#v:maybeToRight), so this `onNothing` can be simplified a bit.

haskell/src/Lorentz/Contracts/Stablecoin.hs
```diff
@@ -407,15 +398,113 @@ stablecoinContract =
...
+  -- but the TZIP-16 spec is explicit about that not being the case.
+  --
+  -- > Unless otherwise-specified, the encoding of the values must be the direct stream
+  -- > of bytes of the data being stored. (...)
+  -- > There is no implicit conversion to Michelson's binary format (PACK) nor
+  -- > quoting mechanism.
+  --
+  -- See: <https://gitlab.com/tzip/tzip/-/blob/eb1da57684599a266334a73babd7ba82dbbbce66/proposals/tzip-16/tzip-16.md#contract-storage>
+  --
+  -- So, instead, we encode it as UTF-8 byte sequences.
```
sras, line 430: This is something I wondered about while going through the TZIP-16. IIRC It does not mention an encoding here, and just provide some samples. I can't think of a reason why they have left it unspecified.

dcastro: Yeah, I was confused by that too.

The examples they give are examples of ASCII encoding:

> For instance, an URI starting with http: will start with the 5 bytes 0x687474703a (h is 0x68, t is 0x74, etc.).

I think they're assuming URIs can only contain ascii characters (which I don't think is [true](https://en.wikipedia.org/wiki/Internationalized_domain_name)), and so they're implying it should be ASCII-encoded (which is a subset of UTF-8-encoding).

I agree, it should be explicit in the TZIP.

The same goes for the metadata JSON, which definitely can contain non-ascii unicode characters, and therefore could be encoded as UTF-8, UTF-16, ...

haskell/src/Stablecoin/Util.hs
dcastro: Ah you're right! I completely missed that because it's all in one :so-big: line 😅

I'll delete `readContract'` from stablecoin, and make an MR in morley to spread the error message across multiple lines to make it easier to read.

image

haskell/test/Lorentz/Contracts/Test/Permit.hs
dcastro: Ah, forgot about that! I deleted `onNothing`

dcastro:
> Btw, xrefcheck reports some errors now, apparently they are caused by changed in this PR (see CI).

Fixed

haskell/src/Lorentz/Contracts/Spec/TZIP16Interface.hs
```diff
@@ -0,0 +1,241 @@
...
+      evalEither (Left err) = fail (pretty err)
+
+----------------------------------------------------------------------------
+-- Handling views
+----------------------------------------------------------------------------
+
+mkMichelsonStorageView :: [(SomeAnn, Text)] -> Lambda (st, param) ret -> MichelsonStorageView (ToT st) (ToT param) (ToT ret)
+mkMichelsonStorageView anns code =
+  MichelsonStorageView
+    { msvCode = (compileLorentz code)
```
sras, line 180: Redundant brackets?

haskell/src/Lorentz/Contracts/Spec/TZIP16Interface.hs
```diff
@@ -0,0 +1,241 @@
...
+      evalEither :: Either FromExpressionError a -> Parser a
+      evalEither (Right a) = pure a
+      evalEither (Left err) = fail (pretty err)
+
+----------------------------------------------------------------------------
+-- Handling views
+----------------------------------------------------------------------------
+
+mkMichelsonStorageView :: [(SomeAnn, Text)] -> Lambda (st, param) ret -> MichelsonStorageView (ToT st) (ToT param) (ToT ret)
+mkMichelsonStorageView anns code =
```
sras, line 178: Some instructions seems to be forbidden in the view code. Should we check that here?

haskell/src/Lorentz/Contracts/Stablecoin.hs
```diff
@@ -419,3 +410,101 @@ registryContract =
...
+    , mVersion = Just (toText $ showVersion version)
+    , mLicense = Just License { lName = "MIT", lDetails = Nothing }
+    , mAuthors =
+        [ "Serokell <https://serokell.io/>"
+        , "TQ Tezos <https://tqtezos.com/>"
+        ]
+    , mHomepage = Just "https://github.com/tqtezos/stablecoin/"
+    , mSource = Just Source
+        { sLocation = "https://github.com/tqtezos/stablecoin/tree/v" <> toText (showVersion version) <> "/ligo/stablecoin"
+        , sTools = [ "ligo " ] -- TODO: add ligo version
```
sras, line 449: > -- TODO: add ligo version

Not sure if left intentional

haskell/src/Lorentz/Contracts/Stablecoin.hs
```diff
@@ -419,3 +410,101 @@ registryContract =
...
+        }
+    , mInterfaces = [ "TZIP-12", "TZIP-17" ]
+    , mErrors =
+        [ mkError [mt|XTZ_RECEIVED|]               [mt|Contract received a non-zero amount of tokens and should not proceed any further|]
+        , mkError [mt|NOT_CONTRACT_OWNER|]         [mt|Authorized sender is not contract owner|]
+        , mkError [mt|NOT_PENDING_OWNER|]          [mt|Authorized sender is not current contract pending owner|]
+        , mkError [mt|NO_PENDING_OWNER_SET|]       [mt|Thrown when trying to authorize as pending owner whilst is not set for a contract|]
```
sras, line 456: > whilst is not set for a contract

Is something missing here?

haskell/src/Stablecoin/Util.hs
```diff
@@ -0,0 +1,12 @@
+-- SPDX-FileCopyrightText: 2020 TQ Tezos
+-- SPDX-License-Identifier: MIT
+
+module Stablecoin.Util
+  ( aesonOptions
+  ) where
+
+import Data.Aeson
```
sras, line 8: Shoulnd't we use qualified imports here?

sras (approved):
LGTM except for minor issues.

haskell/src/Lorentz/Contracts/Spec/TZIP16Interface.hs
dcastro: Which instructions? :open_mouth:

haskell/src/Lorentz/Contracts/Stablecoin.hs
dcastro: Yea, seems to be :/ I took these messages from the spec, but they need to be revised.

In fact, I think the purpose of having these in the TZIP-16 metadata is so that tools can translate error codes like `NO_PENDING_OWNER_SET`, and show a human-readable message to the user. So these messages should be written with the end-user in mind.

I'll revise them now and submit a new commit.

haskell/src/Lorentz/Contracts/Stablecoin.hs
dcastro: I was thinking we could later parse the ligo compiler version from `nix/sources.json` and insert it here (using TH, for example), but that's not necessary and we can leave that for some other time.

haskell/src/Lorentz/Contracts/Spec/TZIP16Interface.hs
sras: I quite probably have mixed something up lol, but here is what I was referring to

https://gitlab.com/tzip/tzip/-/blob/eb1da57684599a266334a73babd7ba82dbbbce66/proposals/tzip-16/tzip-16.md#michelson-storage-views

```
There are limitations on the contents of the "code" field:

The following instructions must not be used:
AMOUNT,
CREATE_CONTRACT,
SENDER,
SET_DELEGATE,
SOURCE,
and
TRANSFER_TOKENS
```

haskell/src/Lorentz/Contracts/Spec/TZIP16Interface.hs
dcastro: Done

haskell/src/Lorentz/Contracts/Stablecoin.hs
dcastro: Done, please review the new messages

haskell/src/Stablecoin/Util.hs
dcastro: Done

haskell/src/Lorentz/Contracts/Spec/TZIP16Interface.hs
dcastro: Ah good catch, I missed that completely 😅 

This is not a problem in this specific instance, because we know these views do not use those instructions, so I propose not doing it now, and add it as an item to this issue: https://gitlab.com/morley-framework/morley-metadata/-/issues/2

What do you think @gromakovsky ?

haskell/src/Lorentz/Contracts/Stablecoin.hs
sras: Looks good.

haskell/src/Lorentz/Contracts/Spec/TZIP16Interface.hs
gromakovsky: Yes, sounds good to me.

gromakovsky: > add it as an item to this issue: https://gitlab.com/morley-framework/morley-metadata/-/issues/2

So can you please add a comment to that issue or just update its description?

sras: > propose not doing it now, and add it as an item to this issue

Fine by me.

gromakovsky (approved):

Issue #131, opened 2020-10-30
Implement TZIP-16 according to TZIP-12


gromakovsky:
# Clarification and motivation

See https://gitlab.com/tzip/tzip/-/merge_requests/94. Now TZIP-12 (FA2) refers to TZIP-16 and has new requirements. We should update `stablecoin` contract accordingly.

Note: this issue sort of overlaps with #124 because TZIP-17 now also requires TZIP-16 compliance.

# Acceptance criteria

After https://gitlab.com/tzip/tzip/-/merge_requests/94 is merged, `stablcoin` contract should comply with the newest version of FA2.

gromakovsky:
If we should make some updates in FA2 types, it might be better to do #115 before starting this issue, so that we update FA2 types in only one place (in `morley-ledgers`).

sras:
@gromakovsky It appears that we have to implement only the following changes for TZIP-12 compliance.

1. Token_metadata_registery entrypoint and storage field should be removed, and token metadata should be stored in TZIP-16 metadata itself.

2. Implment off chain views for `get_balance`, `total_supply`, `all_tokens`, `is_operator`.

gromakovsky:
> Token_metadata_registery entrypoint and storage field should be removed

Yes.

> token metadata should be stored in TZIP-16 metadata itself.

Apparently there are two options: store it in a `%token_metadata`-annotated `big_map` or as a `token_metadata` off-chain view. The latter should make the storage type smaller, decreasing gas consumption and making operations a bit cheaper. So I think it's preferred. This off-chain view doesn't have to access storage as long as metadata is constant, it can return constant data.

> Impelment off chain views for get_balance, total_supply, all_tokens, is_operator.

These are optional, but I think it's better to provide them anyway. Also if we expose token metadata as an off-chain view, all of these off-chain views will be mandatory (and we'll have to implement `token_metadata` as well).

To sum up, I think we have to make only one smart contract change: remove `token_metadata_registry` and the storage field it uses. On the client side we should extend the metadata storage to contain 5 off-chain views. Besides, we can do some cleanup, e. g. remove the old `metadata.tz` contract.

sras:
>These are optional 

It appears that the spec says

> If present, all of these SHOULD be implemented, at least, as *“Michelson Storage Views”*

Doesn't it mean it should be all or none? So I thought as we are probably implementing `token_metadata` as an off chain view, we should implement others as well.

gromakovsky:
[...]

PR #142, opened 2020-12-14
[#138] Configure contract-metadata storage.

### LINKED ISSUE #138: More flexible metadata

Currently we store all TZIP-16 metadata in the Stablecoin contract. This metadata is quite big, it caused origination operation of the contract to exceed the hard limit. I think there are two possible solutions:
1. TZIP-16 specifies that metadata can be stored in the original contract, in another contract (referred from the original one) or off-chain. Making sure metadata is accessible off-chain seems to be out of scope of `stablecoin-client`, while storing it in a separate contract seems very doable. I think that ideally the user should be able to choose where to store metadata. If it's requested to be stored in place, we originate the contract with all metadata in place. If it's requested to be stored separately, we originate two contracts. If the user wants to store it off-chain, they should put it into the right place and provide URIs.
2. Do deploy in two steps: first originate the contract with no metadata, then put it in the next step.

# Acceptance criteria

At least one of the above solutions is implemented.

We should implement everything mentioned above, unless there is a good reason not to do something.

sras:
## Description

1. Add a `contract-metadata.ligo` contract to hold the contract metadata. 
2. Add options to the deploy command to specify how contract metadata should be set.

Right now two options are added. 

1.  --contract-metadata-remote

If this is given, the contract-metadata contract will be deployed, originated with the contract metadata, and the address will be stored in the metadata bigmap of the main contract, with the URI formated as specified by TZIP-16

2.  --contract-metadata-raw 

If this is provided with a string, the provided string is used as such for the metadata URI.

Apart from this, the part of metadata that have been commented out to save origination cost has been enabled.

## Related issue(s)

Resolves  #138 

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -12,6 +12,12 @@ README when a new version is released. -->
 * [#144](https://github.com/tqtezos/stablecoin/pull/144)
   * Add `total_supply` field to the contract.
 
+* [142](https://github.com/tqtezos/stablecoin/pull/142)
+  * Deploy TZIP-16 metadata to an external contract by default.
+  * Provide deploy options `--contract-metadata-in-place` and
+    `--contract-metadata-off-chain` to let user to embed the metadata
+    in contract or to use a custom metadata offline uri respectievely.
+
 ## 1.4.0
 
 * [#134](https://github.com/tqtezos/stablecoin/pull/134)
--- a/haskell/nettest/Nettest.hs
+++ b/haskell/nettest/Nettest.hs
@@ -67,11 +67,14 @@ scNettestScenario originateTransferlist transferlistType = uncapsNettest $ do
   comment "Originating metadata registry contract"
   mrAddress <- nettestOriginateMetadataRegistry
 
+  comment "Originating contract metadata contract"
+  cmrAddress <- nettestOriginateContractMetadataContract metadataJSON
+
   comment "Originating stablecoin contract"
   scAddress <-
     originateUntypedSimple
       "nettest.Stablecoin"
-      (untypeValue (toVal (mkInitialStorage originationParams mrAddress)))
+      (untypeValue (toVal (mkInitialStorage originationParams mrAddress (Just cmrAddress))))
       (T.convertContract stablecoinContract)
 
   comment "Originating transferlist contract"
@@ -365,7 +368,7 @@ scNettestScenario originateTransferlist transferlistType = uncapsNettest $ do
 
       -- We originate a new contract to make sure the minters list is empty
       scAddress' <- do
-        let str = mkInitialStorage (originationParams { opMinters = mempty }) mrAddress
+        let str = mkInitialStorage (originationParams { opMinters = mempty }) mrAddress (Just cmrAddress)
         originateUntypedSimple "nettest.Stablecoin_for_minter_test" (untypeValue $ toVal str) (T.convertContract stablecoinContract)
 
       let
--- a/haskell/src/Lorentz/Contracts/Stablecoin.hs
+++ b/haskell/src/Lorentz/Contracts/Stablecoin.hs
@@ -3,6 +3,7 @@
 
 module Lorentz.Contracts.Stablecoin
   ( ConfigureMinterParam(..)
+  , ContractMetadataRegistryStorage
   , ChangeMasterMinterParam
   , ChangePauserParam
   , GetCounterParam
@@ -11,8 +12,11 @@ module Lorentz.Contracts.Stablecoin
   , MetadataRegistryStorage'(..)
   , MetadataRegistryStorage
   , MetadataRegistryStorageView
+  , MetadataUri(..)
   , mkMetadataRegistryStorage
+  , mkContractMetadataRegistryStorage
   , defaultMetadataRegistryStorage
+  , defaultContractMetadataStorage
   , MintParams
   , MintParam(..)
   , BurnParams
@@ -22,6 +26,7 @@ module Lorentz.Contracts.Stablecoin
   , Storage
   , StorageView
   , UserPermits(..)
+  , UpdateOperatorData(..)
   , PermitInfo(..)
   , Roles(..)
 
@@ -39,6 +44,7 @@ module Lorentz.Contracts.Stablecoin
   , metadataJSON
 
   -- * Embedded LIGO contracts
+  , contractMetadataContract
   , stablecoinContract
   , registryContract
   ) where
@@ -55,20 +61,29 @@ import Lorentz as L
 import Lorentz.Contracts.Spec.TZIP16Interface
   (Error(..), License(..), Metadata(..), MetadataMap, SomeMichelsonStorageView(..), Source(..),
   ViewImplementation(..))
+import Tezos.Address (formatAddress)
 import Michelson.Test.Import (readContract)
 import qualified Michelson.Typed as T
 import Morley.Client (BigMapId(..))
 import Morley.Metadata (mkMichelsonStorageView)
 import Morley.Micheline (ToExpression(toExpression))
+import Morley.Client (AddressOrAlias(..))
 import qualified Tezos.Crypto as Hash
 
 import qualified Lorentz.Contracts.Spec.FA2Interface as FA2
 import qualified Lorentz.Contracts.Spec.TZIP16Interface as TZ (View(..))
-import Lorentz.Contracts.StablecoinPath (metadataRegistryContractPath, stablecoinPath)
+import Lorentz.Contracts.StablecoinPath
+  (contractMetadataRegistryContractPath, metadataRegistryContractPath, stablecoinPath)
 import Paths_stablecoin (version)
 
 ------------------------------------------------------------------
 -- Parameter
+
+-- | Data needed to add or remove an operator.
+data UpdateOperatorData
+  = AddOperator AddressOrAlias
+  | RemoveOperator AddressOrAlias
+  deriving stock Show
 
 data ConfigureMinterParam = ConfigureMinterParam
  { cmpMinter :: Address
@@ -316,6 +331,14 @@ deriving anyclass instance IsoValue StorageView
 minterLimit :: Int
 minterLimit = 12
 
+data ContractMetadataRegistryStorage' big_map = ContractMetadataRegistryStorage
+  { cmrsDummyField :: () -- Dummy field to match the ligo contract's storage.
+  , cmrsMetadata:: MetadataMap big_map
+  }
+  deriving stock Generic
+
+type ContractMetadataRegistryStorage = ContractMetadataRegistryStorage' BigMap
+
 data MetadataRegistryStorage' big_map = MetadataRegistryStorage
   { mrsDummyField :: ()
   , mrsTokenMetadata :: big_map FA2.TokenId FA2.TokenMetadata
@@ -325,11 +348,21 @@ data MetadataRegistryStorage' big_map = MetadataRegistryStorage
 type MetadataRegistryStorage = MetadataRegistryStorage' BigMap
 type MetadataRegistryStorageView = MetadataRegistryStorage' BigMapId
 
-deriving stock instance Show (MetadataRegistryStorage)
-deriving stock instance Show (MetadataRegistryStorageView)
+deriving stock instance Show MetadataRegistryStorage
+deriving stock instance Show MetadataRegistryStorageView
 
-deriving anyclass instance IsoValue (MetadataRegistryStorage)
-deriving anyclass instance IsoValue (MetadataRegistryStorageView)
+deriving anyclass instance IsoValue MetadataRegistryStorage
+deriving anyclass instance IsoValue MetadataRegistryStorageView
+deriving anyclass instance IsoValue ContractMetadataRegistryStorage
+
+-- | Construct the storage for the Contract Metadata contract.
+mkContractMetadataRegistryStorage
+  :: MetadataMap BigMap
+  -> ContractMetadataRegistryStorage
+mkContractMetadataRegistryStorage m = ContractMetadataRegistryStorage
+  { cmrsDummyField = ()
+  , cmrsMetadata = m
+  }
 
 -- | Construct the storage for the Metadata Registry contract.
 mkMetadataRegistryStorage :: MText -> MText -> Natural -> MetadataRegistryStorage
@@ -355,6 +388,14 @@ defaultMetadataRegistryStorage =
     [mt|TEST|]
     [mt|Test|]
     8
+
+-- | The default storage for the contract metadata storage
+-- with some hardcoded token metadata.
+-- Useful for testing.
+defaultContractMetadataStorage :: ContractMetadataRegistryStorage
+defaultContractMetadataStorage =
+  mkContractMetadataRegistryStorage
+    (BigMap $ Map.fromList [([mt|TEST|], "TEST"), ([mt|Test|], "Test")])
 
 -- This empty splice lets us workaround the GHC stage restriction, and refer to `Storage`
 -- in the TH splices below.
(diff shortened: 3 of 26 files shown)


ligo/stablecoin/contract-metadata.ligo
```diff
@@ -0,0 +1,21 @@
+// SPDX-FileCopyrightText: 2020 TQ Tezos
+// SPDX-License-Identifier: MIT
+
+#include "spec.ligo"
```
gromakovsky, line 4: 1. Are we using anything from `spec.ligo`?
2. Maybe let's call this module `tzip16-metadata`? Now we have `metadata.tz` and `contract-metadata.tz`, it's not obvious what the difference between them is.

gromakovsky:
I started reviewing, but didn't review much. There are some fixup commits, you can squash them, as well as other commits fixing previous commits. We've just moved TZIP-16 code into `morley-metadata`, see https://github.com/tqtezos/stablecoin/pull/143, it brought some conflicts. Maybe it would be better to merge this PR first :thinking: 
Anyway, it looks like these conflicts should be easy to resolve. I am not sure if we want to port anything added here into `morley-metadata`, cc @dcastro

sras:
@gromakovsky @dcastro Have cleaned up the commits and expanded messages.

ligo/stablecoin/contract-metadata.ligo
sras: Removed spec.ligo import and renamed the contract.

gromakovsky:
Btw, maybe it's just me, but `--contract-metadata-remote` and `--contract-metadata-raw` names sound somewhat confusing to me. How about `--contract-metadata-on-chain` and `--contract-metadata-off-chain`?

sras:
>How about --contract-metadata-on-chain and --contract-metadata-off-chain

@gromakovsky But metadata embedded in the main contract itself is also on-chain, right? `--contract-metadata-off-chain` appears fine.

gromakovsky:
Oh, right, I just didn't realize that by default we put metadata into the current contract. Then `--contract-metadata-on-chain` is bad indeed. But does it work when we originate the contract with the whole metadata in one operation? I remember we had an issue with operation size and had to comment out some stuff which is uncommented now.

Even if originating with whole metadata in one operation works now, I am afraid it may stop working when we add more functionality (e. g. `total_supply` added in another PR). So I propose to make `RemoteContract` method the default one. If/when 16384 bytes stops being enough for the whole contract+metadata, we can remove some optional metadata fields and the default way will still include all the fields.

Then we can call the second option `--contract-metadata-in-place`, for example. Does it sound ok?

haskell/src/Stablecoin/Client/Parser.hs
```diff
@@ -278,8 +281,31 @@ clientArgsRawParser = Opt.subparser $
...
           (#help .! "Number of seconds it takes for a permit to expire")
+
+      -- Where to place contract metadata, while keeping it embedded in
+      -- contract being the default
+      dcoContractMetadata <- fromMaybe OpCurrentContract <$> contractMetadataOptParser
       pure $ DeployContractOptions {..}
 
+    contractMetadataOptParser :: Opt.Parser (Maybe ContractMetadataOptions)
+    contractMetadataOptParser =
+        optional (rawOptParser <|> remoteContractOptParser)
```
dcastro: When writing CLI tools, it's good practice to fail when an unexpected/unrecognized option is passed.

The general idea is: it would be very bad if, instead of typing `delete --folder "data"`, I wore to make a typo and typed `delete --flder "data"`, the tool silently ignored the `--flder "data"` option and ended up deleting the current folder's contents instead. So it's best to be really strict and reject all typos/unexpected options.

In our case, there's an unwritten rule that `--contract-metadata-raw` and `contract-metadata-remote` are mutually exclusive: you can use one, or the other, or neither, but never both.

Using `<|>` means if both options are passed, we'll silently ignore the second.
Instead, the parser should fail.

sras:
@gromakovsky Have updated the deploy options.

haskell/src/Stablecoin/Client/Parser.hs
```diff
@@ -278,8 +282,29 @@ clientArgsRawParser = Opt.subparser $
...
+          Nothing
+          (#name .! "contract-metadata-off-chain")
+          (#help .! "Use the provided off-chain URI for contract metadata URI"))
+
+        inplaceOptParser :: Opt.Parser ContractMetadataOptions
+        inplaceOptParser =
+          Opt.flag' OpCurrentContract
+            (Opt.long "contract-metadata-in-place"
+            <> Opt.help
+              "Embedd the metadata for the contract in the contract's storage itself.")
```
dcastro, line 306:
```suggestion
              "Embed the metadata for the contract in the contract's storage itself.")
```

ChangeLog.md
```diff
@@ -8,6 +8,13 @@ SPDX-License-Identifier: MIT
 <!-- Prepend new entries here -->
 <!-- Don't forget to update the gas/transaction costs tables in the
 README when a new version is released. -->
+
+*[142](https://github.com/tqtezos/stablecoin/pull/142)
```
dcastro, line 12:
```suggestion
* [142](https://github.com/tqtezos/stablecoin/pull/142)
```

dcastro (changes requested):

sras:
@dcastro Your suggestions make sense. I ll make the changes and update.

haskell/src/Stablecoin/Client/Contract.hs
```diff
@@ -2,21 +2,37 @@
...
-import Michelson.Text (MText)
 import Morley.Client (AddressOrAlias)
-import Tezos.Address (Address)
 
-import Lorentz.Contracts.Stablecoin (Expiry, Roles(..), Storage, Storage'(..), metadataMap)
+import Lorentz as L
+import Lorentz.Contracts.Spec.TZIP16Interface (MetadataMap)
+import Lorentz.Contracts.Stablecoin (Expiry, Roles(..), Storage, Storage'(..))
+
+data ContractMetadataOptions = OpCurrentContract | OpRemoteContract | OpRaw Text
```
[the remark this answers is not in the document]
sras: Originally I didn't put it in Parser module because it appeared to cause circular imports. But surprisingly, I was able to solve it here by moviing something around. So have moved it to Parser module now and have added comments.

haskell/src/Stablecoin/Client/Parser.hs
sras: When I wrote this, I just assumed that the `opt-parse` library handles this. But now when I look it up in manual, it appears to say this under the 'Alternative' section. I tested it and it indeed seem to reject if both arguments are provided.

> Now --file "foo.txt" will be parsed as FileInput "foo.txt", --stdin will be parsed as StdInput, but a command line containing both options, will be rejected.

haskell/src/Stablecoin/Client/Impl.hs
```diff
@@ -74,7 +76,7 @@ data UpdateOperatorData
 --
 -- Saves the contract with the given alias.
 -- If the given alias already exists, nothing happens.
-deploy :: "sender" :! AddressOrAlias -> AliasHint -> InitialStorageData AddressOrAlias -> MorleyClientM (Text, Address, Address)
+deploy :: "sender" :! AddressOrAlias -> AliasHint -> InitialStorageData AddressOrAlias -> MorleyClientM (Text, Address, Address, Maybe Address)
```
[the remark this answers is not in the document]
sras: Have changed

haskell/src/Lorentz/Contracts/Stablecoin.hs
```diff
@@ -314,6 +321,14 @@ deriving anyclass instance IsoValue StorageView
 minterLimit :: Int
 minterLimit = 12
 
+data ContractMetadataRegistryStorage' big_map = ContractMetadataRegistryStorage
+  { cmrsDummyField :: ()
```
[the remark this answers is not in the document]
sras: Have added the comments. When writing this I spotted another mistake with the field name of the actual metadata. It was given as `metadataJSON` but it seems that should be `metadata` as per TZIP-16. Have corrected it as well.

haskell/src/Stablecoin/Client/Impl.hs
```diff
@@ -97,24 +99,50 @@ deploy (arg #sender -> sender) alias initialStorageData = do
         registryContract
         (toVal registryStorage)
 
+  -- Make the contract metadata bigmap
+  let contractMetadataField = [mt|metadataJSON|]
```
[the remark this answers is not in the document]
sras: I originally thought we might give the user to specify the fieldname in the CLI option, and that was why it is part of the constructor. But I ended up not implementing that option. So I have removed that field from the constructor and have defined inside that function you have linked.

haskell/src/Lorentz/Contracts/Stablecoin.hs
```diff
@@ -430,9 +479,25 @@ metadataMap = BigMap $ Map.fromList
...
-  , ([mt|metadataJSON|], BSL.toStrict (J.encode metadataJSON))
-  ]
+  case mdata of
+    CurrentContract jfield md ->
+      [ (mempty, encodeUtf8 @Text $ "tezos-storage:" <> (toText jfield))
+      , (jfield, BSL.toStrict (J.encode md))
+      ]
+    RemoteContract addr jfield ->
+      [ (mempty, encodeUtf8 @Text $
+          "tezos-storage://" <> (formatAddress addr) <> "/" <> (toText jfield))
```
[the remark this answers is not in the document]
sras: I have now removed the field from the constructor and have defined it here. AFAIU it takes care of this issue.

haskell/src/Stablecoin/Client/Impl.hs
```diff
@@ -97,24 +99,50 @@ deploy (arg #sender -> sender) alias initialStorageData = do
...
+  contractMetadataUri  <-
+    case isdContractMetadataStorage initialStorageData of
+      -- User wants to store metadata embedded in the contact
+      OpCurrentContract -> pure $ CurrentContract contractMetadataField metadataJSON
+      -- User have stored metadata somewhere and just wants to put the raw uri in metadata bigmap
+      OpRaw url -> pure $ Raw url
+      -- User wants a new contract with metadata to be deployed.
+      OpRemoteContract -> do
+        let mdrStorage = mkContractMetadataRegistryStorage
+              (metadataMap $ CurrentContract contractMetadataField metadataJSON)
```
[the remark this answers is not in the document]
sras: I had some thoughts about it, but in the end it appeared that you are right. We already provide the key in the main contract, so providing key here as well makes it redundant.

haskell/src/Stablecoin/Client/Main.hs
```diff
@@ -59,12 +59,17 @@ mainProgram (ClientArgs _ globalOptions cmd) = do
...
         , isdTokenMetadataRegistry = dcoTokenMetadataRegistry
         , isdDefaultExpiry = dcoDefaultExpiry
+        , isdContractMetadataStorage = dcoContractMetadata
         }
 
       putTextLn "Contract was successfully deployed."
       putTextLn $ "Operation hash: " <> pretty opHash
       putTextLn $ "Contract address: " <> pretty contractAddr
       putTextLn $ "Metadata registry contract address: " <> pretty metadataRegAddr
+      case mCMetadataAddr of
```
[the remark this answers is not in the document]
sras: Ah, I see. Changed.

haskell/src/Stablecoin/Client/Parser.hs
sras: Fixed.

ChangeLog.md
sras: Fixed.

haskell/src/Stablecoin/Client/Impl.hs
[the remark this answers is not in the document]
sras: Changed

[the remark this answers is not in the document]
sras: Changed by adding a boolean flag to `CurrentContract` constructor, and depending on its value include the empty key uri field in the map.

haskell/src/Stablecoin/Client/Parser.hs
dcastro: Oh, didn't know that! That's great then :)

dcastro (approved):

haskell/nettest/Nettest.hs
```diff
@@ -67,11 +67,14 @@ scNettestScenario originateTransferlist transferlistType = uncapsNettest $ do
   comment "Originating metadata registry contract"
   mrAddress <- nettestOriginateMetadataRegistry
 
+  comment "Originating contract metadata contract"
```
gromakovsky, line 70: `contract-metadata` was renamed to `tzip16-metadata`, but in code we still say "contract metadata" in many places. Let's not change it for now, but when we do #131 and remove the legacy interface, we will have only one metadata contract and will call it just "metadata" without "contract" or "tzip-16" prefix.

haskell/test-common/Lorentz/Contracts/Test/Common.hs
```diff
@@ -199,8 +201,8 @@ withOriginated
   -> IntegrationalScenario
 withOriginated fn op tests = fn op >>= tests
 
-mkInitialStorage :: OriginationParams -> Address -> Storage
-mkInitialStorage OriginationParams{..} metadataRegistryAddress =
+mkInitialStorage :: OriginationParams -> Address -> (Maybe Address) -> Storage
```
gromakovsky, line 204:
```suggestion
mkInitialStorage :: OriginationParams -> Address -> Maybe Address -> Storage
```

gromakovsky:
> But does it work when we originate the contract with the whole metadata in one operation? I remember we had an issue with operation size and had to comment out some stuff which is uncommented now.

Ok, I have tested it on my machine and it doesn't work.
```
stablecoin-client: FailureResponse (Request {requestPath = (BaseUrl {baseUrlScheme = Http, baseUrlHost = "delphi.testnet.tezos.serokell.team", baseUrlPort = 8732, baseUrlPath = ""},"/injection/operation"), requestQueryString = fromList [("chain",Just "main")], requestBody = Just ((),application/json;charset=utf-8), requestAccept = fromList [application/json;charset=utf-8,application/json], requestHeaders = fromList [], requestHttpVersion = HTTP/1.1, requestMethod = "POST"}) (Response {responseStatusCode = Status {statusCode = 500, statusMessage = "Internal Server Error"}, responseHeaders = fromList [("content-length","99"),("content-type","application/json")], responseHttpVersion = HTTP/1.1, responseBody = "[{\"kind\":\"permanent\",\"id\":\"node.prevalidation.oversized_operation\",\"size\":16876,\"max_size\":16384}]\n"})
```

So, as I said, I propose to not include optional metadata fields if `--contract-metadata-in-place` is chosen. It's weird to provide `--contract-metadata-in-place` if it doesn't work.

sras:
@gromakovsky I have changed it to drop some errors during deploying. It appears that we can originate without exceeding limits now.

gromakovsky:
@sras I think it's better not to include `errors` at all than to include an incomplete list. AFAIU this field is optional, so we are not obliged to include it. And if we include an incomplete list, one may wrongly assume it's complete.

haskell/test-common/Lorentz/Contracts/Test/Common.hs
gromakovsky: @sras what about this?

gromakovsky:
Hm, apparently in this case `errors` key will still appear in metadata and will be mapped to empty list. With the current metadata interface it seems impossible to avoid `errors` key, but I heard that @dcastro is changing the interface to be builder-like, so that all fields can be omitted. Once that's done, I think we can update this metadata to not include `errors` at all.

gromakovsky (approved):

PR #144, opened 2020-12-15
[#131] Add `total_supply` to the storage and update code/tests to keep it in sync.

sras:
## Description

We have to add `total_supply` field to the contract storage as
required by recent changes to TZIP-12. So we add it in this PR
and update spec, tests, and storage initialization functions.

## Related issue(s)

Resolves #144

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/haskell/test/SMT.hs
+++ b/haskell/test/SMT.hs
@@ -83,7 +83,8 @@ generateContractInputs count = do
          gsPauserPool gsTokenOwnerPool
     operatorsMap = Map.fromList $ zip operators $ repeat ()
     startingStorage = SimpleStorage
-      gsMinterPool ledgerBalances owner masterMinter pauser pendingOwner Nothing operatorsMap isPaused
+      gsMinterPool ledgerBalances owner masterMinter pauser pendingOwner Nothing operatorsMap
+      isPaused (sum $ Map.elems ledgerBalances)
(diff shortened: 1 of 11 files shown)


haskell/test/SMT.hs
```diff
@@ -660,18 +668,21 @@ applyBurn ccSender cs burnparams = do
...
-            then error ("Unexpected burn:" <> show value)
-            else Nothing
-                -- As per FA2, burn should follow the tranfer logic, which allows zero transfer
-                -- from non-existant accounts. So here we should do nothing, instead of throwing an
-                -- error.
-          Just x -> let b = x - value
-            in if b > 0 then Just b else Nothing) ccSender $ ssLedger cs }
-      in Right $ foldr burn cs burnparams
+    then if totalBurn > ssTotalSupply cs
+      then Left NEGATIVE_TOTAL_SUPPLY
```
gromakovsky, line 672: AFAIU if our test generator is written correctly, `NEGATIVE_TOTAL_SUPPLY` should never occur. Isn't it better to use `error` here to indicate that there is a bug in the test itself? For example, few lines below we are doing `error ("Unexpected burn:" <> show value)` when we try to burn non-existing tokens.

sras: True.

gromakovsky (approved):
Okay, now only changelog update is missing.

PR #154, opened 2021-01-08
[#131] Adds some more missing TZIP-16 metadata and amend readme.

sras:
## Description

Since we have added more options w.r.t deploying the contracts
and metadata, we should make it more clear how it is being done in
nettests. So the section is amended.

The TIZP-016 spec describe some top level fields, and we
include most of them in our Metadata, except for the 'description'
field. The 'NEGATIVE_TOTAL_SUPPLY' error is also missing from the list
of errors. So this PR adds them both.

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,13 @@ To measure and collect these numbers:
     tezos-client get balance for nettest
     ```
 1. Run `cd haskell && stack test stablecoin:test:stablecoin-nettest`
+
+   This deploys the FA2 and FA1.2 contracts along with the corresponding
+   metadata contracts. So metadata is deployed to separate contracts and linked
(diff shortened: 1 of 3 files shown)


gromakovsky:
> The tables below show the gas and transaction costs of both versions (FA1.2 and FA2) of the stablecoin contract [v1.6.0](https://github.com/tqtezos/stablecoin/releases/tag/v1.6.0) in Delphinet.

I would write here that origination cost doesn't include metadata origination.

sras:
> I would write here that origination cost doesn't include metadata origination.

@gromakovsky Just added it.

gromakovsky (approved):

PR #155, opened 2021-01-13
[#153] Metadata-related refinements

### LINKED ISSUE #153: CamlCase-ify TZIP-16 keys

There was this change in TZIP-16: https://gitlab.com/tzip/tzip/-/merge_requests/115/diffs
Apparently we haven't updated our code yet (judigng by https://better-call.dev/delphinet/big_map/39468/keys).

# Acceptance criteria

TZIP-16 keys like `michelson-storage-view` are compliant with the latest TZIP-16.
Probably something should be updated in `morley-metadata` first.

dcastro:
## Description

This PR adds various metadata-related refinements, mainly:

1. Upgrade to the latest `morley-metadata` version (which includes the new field naming scheme using camelCase, and validation of a view's code)
2. Include the ligo compiler version in the metadata
3. Fix the names of the FA2 views (`GetBalance` -> `get_balance`)
4. Update the types of the `get_balance` and `is_operator` views
5. Add `--description` CLI option

It's quite a big PR, so I suggest reviewing one commit at a time.

## Related issue(s)

Resolves #153

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.


haskell/stack.yaml
```diff
@@ -13,34 +13,27 @@ extra-deps:
...
   subdirs:
     - code/cleveland
-
-- git:
-    https://gitlab.com/morley-framework/morley.git
-    # ^ CI cannot use ssh, so we use http clone here
-  commit:
-    bdf6fe2f2ef7833dbe7ee65d61cf4b2e4691f365
-  subdirs:
     - code/morley
```
rvem, line 19: Perhaps, it's better to use the version from the hackage. Also, 1.11.1 arrived some time ago, so maybe it makes sense to use it

haskell/src/Lorentz/Contracts/StablecoinPath.hs
```diff
@@ -22,3 +23,7 @@ stablecoinFA1_2Path = "./test/resources/stablecoin.fa1.2.tz"
 -- | The path to the compiled metadata registry.
 metadataRegistryContractPath :: FilePath
 metadataRegistryContractPath = "./test/resources/metadata.tz"
+
+-- | The path to nix's sources.json file.
+sourcesJsonPath :: FilePath
+sourcesJsonPath = "./test/resources/sources.json"
```
rvem, line 29: btw, why isn't it inherited directly from `../nix/sources.json`?

rvem:
Probably not the scope of this issue, but I think it'd nice to have some `bats` tests for the `stablecoin-client` binary

rvem:
Oops, sorry, miss clicked :see_no_evil:

ChangeLog.md
```diff
@@ -9,6 +9,16 @@ SPDX-License-Identifier: MIT
 <!-- Don't forget to update the gas/transaction costs tables in the
 README when a new version is released. -->
 
+* [#155](https://github.com/tqtezos/stablecoin/pull/155)
+  * Added `--description` option to `stablecoin-client`
```
rvem, line 13: Is there an easy way to add a test, that deploys a contract with some description and asserts that the description in the metadata is correct?

haskell/stack.yaml
```diff
@@ -13,34 +13,27 @@ extra-deps:
...
     - code/morley-metadata
     - code/morley-metadata-test
 
 # Stable parts of morley available from Hackage
 - morley-prelude-0.3.0
 - tasty-hunit-compat-0.2
-- lorentz-0.6.2
-- indigo-0.3.0
+- indigo-0.5.0
+- uncaught-exception-0.1.0
```
pasqu4le, line 36: Even tho this is a Serokell library I don't think it falls under "Stable parts of **morley** available from Hackage"

haskell/test-common/Lorentz/Contracts/Test/Common.hs
```diff
@@ -147,11 +147,14 @@ defaultOriginationParams = OriginationParams
...
   , opMinters = mempty
   , opPendingOwner = Nothing
-  , opMetadataUri = CurrentContract (metadataJSON $ Just testFA2TokenMetadata) True
+  , opMetadataUri = CurrentContract metadata True
   , opTransferlistContract = Nothing
   , opDefaultExpiry = 1000
   , opPermits = mempty
   }
+  where
+    metadata = either (error . show) id $ metadataJSON (Just testFA2TokenMetadata)
```
pasqu4le: In nettests you usually use `failure . build`, so why not use `fmt` here as well:
```suggestion
    metadata = either (error . pretty) id $ metadataJSON (Just testFA2TokenMetadata)
```

haskell/src/Lorentz/Contracts/Stablecoin/Types.hs
```diff
@@ -127,33 +127,47 @@ deriving anyclass instance HasAnnotation SetExpiryParam
 instance Buildable SetExpiryParam where
   build = genericF
 
-instance HasAnnotation FA2.Parameter where
+-- | Similar to 'FA2.Parameter' from @morley-ledgers@, but with a layout compatible with
+-- the ligo implementation
+data FA2Parameter
```
pasqu4le, line 132: Oh I thought (referring to the whole commit) that this would have been unnecessary after [this recent MR](https://gitlab.com/morley-framework/morley-ledgers/-/merge_requests/30).

Is this only because of the HLS you mentioned then? Like, would it compile fine?

haskell/nettest/FA1_2Comparison.hs
```diff
@@ -57,7 +57,7 @@ fa1_2ComparisonScenario = uncapsNettest $ do
         { sDefaultExpiry = 1000
         , sLedger = BigMap balances
         , sMintingAllowances = mempty
-        , sIsPaused = False
+        , sPaused = False
```
pasqu4le, line 60: Writing this here for lack of a better place, in this commit (89f6743fe548804e7c1d3233f95d770a95d1e042) the message should be corrected, towards the end it says:
> We do, however, have to be careful and make sure the haskell field names the ligo field names,

which I suppose should be:
> We do, however, have to be careful and make sure **that** the haskell field names **match** the ligo field names,

(maybe the `that` is unnecessary tho)

.gitignore
```diff
@@ -17,3 +17,4 @@ TAGS
 /haskell/test/resources/stablecoin.tz
 /haskell/test/resources/stablecoin.fa1.2.tz
 /haskell/test/resources/metadata.tz
+/haskell/test/resources/sources.json
```
pasqu4le, line 20: Here too the comment of this "group" doesn't match this addition:
> We also ignore them in `haskell/test/resources/` because they are symlinked for tests.

I think a new comment to explain why we need this (or rather why it is there) would be the best

haskell/package.yaml
```diff
@@ -45,6 +45,8 @@ library:
     - bytestring
     - uncaught-exception
     - servant-client
+    - template-haskell
+    - directory
```
pasqu4le, line 49: I was going to mention how `directory` is not in alphabetical order, but oh boy it's not the only one :sweat_smile: 

Would you mind please fixing that (a fixup to any commit that changes this file is ok IMO).

haskell/src/Stablecoin/Client/Parser.hs
```diff
@@ -168,9 +168,11 @@ data GetMintingAllowanceOptions = GetMintingAllowanceOptions
 
 data ContractMetadataOptions
   = OpCurrentContract -- ^ Indicate that we should store metadata in the main contract itself.
+      (Maybe Text) -- ^ An optional description of the contract to add to the contract's metadata.
   | OpRemoteContract -- ^ Indicate we should originate a dedicated contract to store metadata.
+      (Maybe Text) -- ^ An optional description of the contract to add to the contract's metadata.
```
pasqu4le, line 173: I suppose `add` is not quite correct here, because if non is specified the default one will be used?

Maybe replace it with:
```suggestion
      (Maybe Text) -- ^ An optional contract description to replace the default one in its metadata.
  | OpRemoteContract -- ^ Indicate we should originate a dedicated contract to store metadata.
      (Maybe Text) -- ^ An optional contract description to replace the default one in its metadata.
```

haskell/src/Stablecoin/Client/Parser.hs
```diff
@@ -300,11 +302,27 @@ clientArgsRawParser = Opt.subparser $
...
+
+        remoteContractOptParser :: Opt.Parser ContractMetadataOptions
+        remoteContractOptParser = OpRemoteContract <$> contractDescriptionParser
+
+        contractDescriptionParser :: Opt.Parser (Maybe Text)
+        contractDescriptionParser =
+          optional $ mkCLOptionParser @Text
+            Nothing
+            (#name .! "description")
+            (#help .! "An optional description of the contract to add to the contract's metadata. \
```
pasqu4le, line 322: Same comment as before on `add`, although here is less important given the explanation following.

Could be rephrased in a different way as well.

pasqu4le (commented):
Only a few nipicks really, otherwise LGTM :+1:

haskell/src/Lorentz/Contracts/StablecoinPath.hs
dcastro: Correct me if I'm wrong, but I think when you're building the haskell project using nix with `nix build -L -f. lib`, you can't access any files outside the `./haskell` folder.

haskell/src/Lorentz/Contracts/Stablecoin/Types.hs
dcastro: Yeah, it compiles fine, it's just that it breaks HLS :/

haskell/src/Lorentz/Contracts/StablecoinPath.hs
rvem: hmm, yes, you're right

haskell/src/Lorentz/Contracts/Stablecoin/Types.hs
pasqu4le: Ok I guess this is fine anyway, both solutions are kinda meh, if possible I guess we should look into how to avoid the problem in the first place :thinking:

haskell/package.yaml
dcastro: Done

.gitignore
dcastro: Done

haskell/nettest/FA1_2Comparison.hs
dcastro: Good catch! Done

haskell/test-common/Lorentz/Contracts/Test/Common.hs
dcastro: Done

haskell/stack.yaml
dcastro: Good point! I guess the same applies to `tasty-hunit-compat`, so I've moved that too.

haskell/src/Stablecoin/Client/Parser.hs
dcastro: I personally prefer to use "add" here. "Replace" implies something was already there, which you removed and replaced with something else.
In this case, I'm adding the given description to the contract's metadata, or, if none is given, I'm adding a default description.

I updated the comment to mention what happens when no description is given.

haskell/src/Lorentz/Contracts/Stablecoin/Types.hs
dcastro: I started a discussion [here](https://serokell.slack.com/archives/CF2UVU8RX/p1610462545046100), I'd appreciate your input :)

haskell/stack.yaml
dcastro: Done

haskell/src/Stablecoin/Client/Parser.hs
pasqu4le: Good point, sounds good to me.

pasqu4le (approved):
@dcastro CI is unhappy because `nix/source.json` needs to be updated AFAIK, you'll need a `hackage.nix` there that has `morley-1.11.1`, this should work:
```
    "hackage.nix": {
        "branch": "master",
        "description": "Automatically generated Nix expressions for Hackage",
        "homepage": "",
        "owner": "input-output-hk",
        "repo": "hackage.nix",
        "rev": "3c6b6e040a6c65740a659c0ccefb501155acf086",
        "sha256": "1mdawpvw2nnlr9jqmxwv5m3p86gcbzzbjiyzpims23zy3wacr8c0",
        "type": "tarball",
        "url": "https://github.com/input-output-hk/hackage.nix/archive/3c6b6e040a6c65740a659c0ccefb501155acf086.tar.gz",
```

Everything else LGTM :ok_hand:

rvem (approved):

ChangeLog.md
dcastro: I tried adding this to `StablecoinClientTest` (and modifying `StablecoinImpl` to pass the `--description ` to `stablecoin-client`), but I just remembered that we can't read a contract's metadata yet, not until morley#346 is completed :/

I've created a new issue so we can do this later (#158) and pushed my work to `diogo/#158-test-metadata`.

PR #160, opened 2021-02-01
Add 008 network tests 

rvem:
## Description
Problem: New protocol arrived.

Solution: Run network tests on 008 local-chain and edonet.

Based on https://gitlab.com/morley-framework/morley/-/merge_requests/760

## Related issue(s)

https://issues.serokell.io/issue/TM-491

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/haskell/README.md
+++ b/haskell/README.md
@@ -19,6 +19,7 @@ Otherwise, you should prefix all commands with `stack run` and put `--` after `s
 
 The help message should be sufficiently descriptive, however, there are some caveats:
 1. [`tezos-client`](http://tezos.gitlab.io/introduction/howtoget.html) executable must be available and is assumed to be in `$PATH` by default.
+   Note that only `tezos-client` versions v8.0 or above are supported.
 1. Node data (address, port and whether to use TLS) is taken from `tezos-client` config by default.
 Make sure it points to an active node in the network where you want to submit your operations.
 You can override this data using the respective options of `stablecoin-client`.
--- a/haskell/nettest/Nettest.hs
+++ b/haskell/nettest/Nettest.hs
@@ -107,7 +107,7 @@ scNettestScenario originateTransferlist transferlistType = uncapsNettest $ do
       (#amount .! value)
 
     callTransferWithOperator op from to value =
-      callFrom (AddressResolved op) sc (Call @"Transfer") (tp from to value)
+      withSender (AddressResolved op) $ call sc (Call @"Transfer") (tp from to value)
 
     callTransfer = join callTransferWithOperator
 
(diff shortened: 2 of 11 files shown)


rvem:
:confused: , okay looks like morley-metadata should be updated first

rvem:
Also, looks like we need to export `SenderCap` related stuff from https://gitlab.com/morley-framework/morley/-/blob/master/code/cleveland/src/Morley/Nettest/Caps.hs to reuse them in stablecoin-client

haskell/stack.yaml
```diff
@@ -13,7 +13,7 @@ extra-deps:
     https://gitlab.com/morley-framework/morley.git
     # ^ CI cannot use ssh, so we use http clone here
   commit:
-    9e69d52741e456bd0976a1ec4c45cc4c43a11722 # master, morley-1.11.1
+    56d74ababeda9aba8e194783fc5e395fa7a6b69a # morley-1.12.0
```
gromakovsky, line 16: I don't know if we still support `tezos-client-7.5` after this change, but in https://github.com/tqtezos/stablecoin/blob/master/haskell/README.md we have
> tezos-client executable must be available and is assumed to be in $PATH by default.

and I think it should be updated to require 8.0 or newer

rvem: Updated README

haskell/README.md
```diff
@@ -19,6 +19,7 @@ Otherwise, you should prefix all commands with `stack run` and put `--` after `s
 
 The help message should be sufficiently descriptive, however, there are some caveats:
 1. [`tezos-client`](http://tezos.gitlab.io/introduction/howtoget.html) executable must be available and is assumed to be in `$PATH` by default.
+   Note that only `tezos-client` from version v8.0 is supported.
```
pasqu4le, line 22: Nitpick rephrasing.
```suggestion
   Note that only `tezos-client` versions v8.0 or above are supported.
```

haskell/nettest/Nettest.hs
```diff
@@ -107,16 +107,17 @@ scNettestScenario originateTransferlist transferlistType = uncapsNettest $ do
...
       -> capsM ()
     callTransfers = mapM_ $ \(from@(arg #from_ -> from_), destinations) ->
-      callFrom
-        (AddressResolved from_)
+      withSender
+        (AddressResolved from_) $
+        call
         sc
         (Call @"Transfer")
         (constructTransfersFromSender from destinations)
```
pasqu4le, line 123:
```suggestion
      withSender (AddressResolved from_) $
        call sc (Call @"Transfer") (constructTransfersFromSender from destinations)
```

haskell/nettest/Nettest.hs
```diff
@@ -126,24 +127,27 @@ scNettestScenario originateTransferlist transferlistType = uncapsNettest $ do
...
         sc'
         (Call @"Configure_minter")
         (ConfigureMinterParam for expectedAllowance newAllowance)
 
     removeMinter :: Address -> Address -> capsM ()
     removeMinter from whom =
-      callFrom
-        (AddressResolved from)
+      withSender
+        (AddressResolved from) $
```
pasqu4le, line 140: So... there's a bunch more cases of the two suggestions above in this file :sweat_smile: 

The suggestion is to move the first argument of `withSender` to its line and add indentation to the arguments of `call`.
Note that in many cases where now `call`  can fit with its arguments on a single line of ~80 and I think we should do that where possible.

haskell/nettest/Nettest.hs
```diff
@@ -126,24 +127,27 @@ scNettestScenario originateTransferlist transferlistType = uncapsNettest $ do
...
     configureMinter' :: TAddress Parameter -> Address -> Address -> Maybe Natural -> Natural -> capsM ()
     configureMinter' sc' from for expectedAllowance newAllowance =
-      callFrom
-        (AddressResolved from)
+      withSender
+        (AddressResolved from) $
+        call
         sc'
         (Call @"Configure_minter")
         (ConfigureMinterParam for expectedAllowance newAllowance)
```
pasqu4le, line 135:
```suggestion
      withSender (AddressResolved from) $
        call
          sc'
          (Call @"Configure_minter")
          (ConfigureMinterParam for expectedAllowance newAllowance)
```

pasqu4le (commented):
Minor suggestions.

haskell/nettest/Nettest.hs
rvem: Looks like I've updated everything now

pasqu4le:
Ok one last thing to do after [the morley MR](https://gitlab.com/morley-framework/morley/-/merge_requests/760) gets merged:

in `Stablecoin.Client.Parser` we re-define `addressOrAliasOption`, which should no longer be necessary after that, so let's removeit and import the function from `morley-client` instead.

Aside from that all LGTM, so feel free to squash already when you update the `stack.yaml` commit again.

pasqu4le (approved):
All LGTM now :ok_hand:

PR #172, opened 2021-05-18
[TM-546] Remove edo2net tests

ixahedron:
## Description

Problem: protocol 008 has become obsolete.

Solution: remove tests for 008 local chain and edo2net from the CI.

## Related issue(s)

Resolves part of TM-546

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -70,12 +70,12 @@ steps:
      - nix build -L -f .. test
      - ./result/bin/stablecoin-test
 
- - label: nettest-local-chain-008
-   key: nettest-local-chain-008
+ - label: nettest-local-chain-009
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -146,8 +119,8 @@ steps:
    if: *not_scheduled
    depends_on:
```
rvem, line 120: hmm, why packaging depends on nettests? :thinking:

rvem (approved):

.buildkite/pipeline.yml
ixahedron: It's buildkite dependencies, meaning he packaging step should only run after nettests are run to ensure they pass, I assume.

PR #179, opened 2021-10-14
[#136] Remove coercions between 'Alias' and 'AliasHint'

### LINKED ISSUE #136: Review usages of `Alias`/`AliasHint`

In #135, we updated our morley dependencies.
Due to recent changes in `cleveland`/`morley-client`, the `Alias` type has now been split in two newtypes: `Alias`/`AliasHint`.
We should review the places where we're using these types, and make sure we're using the correct one everywhere.

See comment here: <https://github.com/tqtezos/stablecoin/blob/d0e90637669dc464a45453ad26aee4e1dfbe7951/haskell/src/Stablecoin/Client/Main.hs#L47-L50>

# Acceptance criteria

* We don't `coerce` between those two data types
* We use them correctly
* We don't assume that no alias prefix was set via a CLI option

DMozhevitin:
Problem: We use `coerce` to convert `AliasHint` to alias in
`stablecoin-client`. This coersion is safe only when no alias
prefix was set via CLI option.

Solution: Added prefix accounting when converting from `AliasHint`
to `Alias`, hence the conversion became safe even if the prefix
option was set.

## Related issue(s)

Resolves #136

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

### DIFF
--- a/haskell/src/Stablecoin/Client/Main.hs
+++ b/haskell/src/Stablecoin/Client/Main.hs
@@ -7,7 +7,6 @@ module Stablecoin.Client.Main
   ) where
 
 import Control.Exception.Uncaught (displayUncaughtException)
-import Data.Coerce (coerce)
 import Data.Map (lookup)
 import Fmt (pretty)
 import Michelson.Text (mt)
@@ -15,6 +14,8 @@ import Morley.Client
   (AddressOrAlias(AddressAlias), Alias(..), AliasHint(..), MorleyClientM,
   TezosClientError(UnknownAddressAlias), mkMorleyClientEnv, rememberContract, runMorleyClientM)
 import Morley.Client.TezosClient (resolveAddress)
+import Morley.Client.TezosClient.Impl (prefixName)
+import Morley.Client.TezosClient.Types (HasTezosClientEnv(..), TezosClientEnv(..))
 import qualified Options.Applicative as Opt
 import Util.Named ((.!))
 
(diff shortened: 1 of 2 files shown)


haskell/src/Stablecoin/Client/Main.hs
```diff
@@ -44,13 +45,13 @@ mainProgram :: ClientArgs -> MorleyClientM ()
 mainProgram (ClientArgs _ globalOptions cmd) = do
   case cmd of
     CmdDeployContract DeployContractOptions {..} -> do
-      let contractAlias = AliasHint "stablecoin"
-      -- TODO: we're using coerce in a few places to convert `AliasHint` to `Alias`
-      -- where necessary, but we shouldn't be doing this because it assumes no alias
-      -- prefix option was passed in via the CLI.
-      aliasAlreadyExists <- checkIfAliasExists (coerce contractAlias)
+      prefix <- tceAliasPrefix <$> view tezosClientEnvL
```
pasqu4le, line 48: a bit sad that this is tied to the logic inside morley client itself, but I suppose that there is little we can do if we want to print the full alias in the messages to the user.

There is some space to remove duplication in morley client itself tho, I'll open an issue for this there.

DMozhevitin: > a bit sad that this is tied to the logic inside morley client itself

Maybe it's worth adding a convenient way to get full alias with prefix to `morley-client` ? :thinking: 

Btw, I noticed a piece of code which moved to a separate function in `morley-client` (namely `setupMoneybagAddress`), and we can use it there when we update `morley` version. Should I left a `TODO` for this purpose, or it's better to add a comment to issue related to bumping `morley` dependency (but I can't find it :sweat_smile:)

PR #188, opened 2022-03-15
[Chore] Set last Ligo version tested on a release

pasqu4le:
## Description

Problem: the ligo README currently points at a rolling release version,
which is not very convenient to get in case a new version breaks
compatibility. Moreover, there are now issues with fetching such an old
ligo release from source.

Solution: set the last tested working ligo version to 0.10 and
update the nix sources to get ligo as a static binary.

## Related issue(s)

None

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.


lierdakil:
Uh... the build fails because of some TH pulling in the ligo version.

pasqu4le:
yep, just noticed, it should be fixed now

PR #189, opened 2022-03-17
[Chore] Update copyright holders and licensing info

pasqu4le:
## Description

Copyright holders and licensing info need to be updated and this MR does that.

See commit messages and related issue for more details.

_For reference on how the header replacements were made see [this `morley` comment](https://gitlab.com/morley-framework/morley/-/merge_requests/1085#note_876214999)._

## Related issue(s)

[TM-592](https://issues.serokell.io/issue/TM-592)

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.


LICENSES/LicenseRef-MIT-TQ.txt
```diff
@@ -1,19 +0,0 @@
-MIT License Copyright (c) 2020 Tocqueville Group
```
rvem, line 1: In some PRs, e.g. [here](https://github.com/serokell/tezos-packaging/pull/406), the old LicenseRef file wasn't removed, while here it was removed. Shouldn't it be consistent?

rvem: Oh, right, because in this project this license is no longer referenced, while in the other it still might be referenced

rvem (approved):

LICENSES/LicenseRef-MIT-TQ.txt
pasqu4le: That's correct. Alternatively we could have changed the content of this file, but I didn't do so because it has `TQ` in the actual file name.

PR #198, opened 2022-09-21
[Chore] Add kathmandu protocol support for CI

krendelhoff2:
## Description

This PR upgrades the morley dependencies and adds support for the kathmandu protocol.

I also took the opportunity to clean up the tests a bit.

Follow-up issues:
1. https://github.com/tqtezos/stablecoin/issues/200, depends on:
    * https://gitlab.com/morley-framework/morley/-/issues/896
2. https://github.com/tqtezos/stablecoin/issues/201, depends on:
    * https://gitlab.com/morley-framework/morley/-/issues/889
    * https://gitlab.com/morley-framework/morley-metadata/-/issues/18
    * https://gitlab.com/morley-framework/morley/-/issues/895

## Related issue(s)

None

- Tests
        silently reappearing again.

- Documentation
    - Haddock
        not essential.

#### ✓ Release Checklist

- [x] I updated the [Gas / Transaction costs](https://github.com/tqtezos/stablecoin/tree/master/README.md#gas--transaction-costs) table in the README.
- [x] I updated the version number in `package.yaml`.
- [x] I updated the [changelog](https://github.com/tqtezos/stablecoin/tree/master/ChangeLog.md) and moved everything
      under the "Unreleased" section to a new section for this release version.
- [ ] (After merging) I created a new entry in the [releases](https://github.com/tqtezos/stablecoin/releases) page,
      with a summary of all user-facing changes.
    * I made sure a tag was created using the format `vX.Y.Z`
    * I attached the following artifacts from [buildkite](https://buildkite.com/serokell/stablecoin)
      * `stablecoin-client`
      * `stablecoin.fa1.2.tz`
      * `stablecoin.tz`


dcastro:
The build seems to be broken in master, @krendelhoff please change the `sha256` of `ligo` in sources.json to `167lq0kjxq1m48sp7cpwka9c3jgypryww5xwkp3j065rrw4y9di3`

haskell/test/Lorentz/Contracts/Nettest/StablecoinClientTest.hs
```diff
@@ -38,9 +37,9 @@ test_stablecoinClientScenario =
 -- | Check that all the `stablecoin-client` commands work.
 stablecoinClientScenario :: StablecoinT m ()
 stablecoinClientScenario = do
-  Moneybag originator <- asks $ ccMoneybag . SC.scCleveland
-
   comment "Creating roles"
+  originator <- newAddress "originator"
+  transfer originator [tz|5|]
```
lierdakil, line 42:
```haskell
originator <- refillable $ newAddress "originator"
```
?

The downside refillables are somewhat slower. The upside is they don't require hard-coded amounts.

ChangeLog.md
```diff
@@ -9,6 +9,9 @@ SPDX-License-Identifier: MIT
 <!-- Don't forget to update the gas/transaction costs tables in the
 README when a new version is released. -->
```
pasqu4le, line 11: I think that it would be nice to make a new release right after adding kathmandu support, so IMO it would be nice to update the costs as well as bump the version number in this PR

haskell/test/Lorentz/Contracts/Nettest/StablecoinClientTest.hs
```diff
@@ -35,45 +35,40 @@ test_stablecoinClientScenario =
...
-  (transferlistAddr, transferlist) <- createRole "transferlist"
+  originator <- refillable $ newAddress "originator"
+  masterMinter <- newAddress "master-minter"
+  pauser <- newAddress "pauser"
+  contractOwner <- newAddress "contract-owner"
+  minter <- newAddress "minter"
+  transferlist <-
+    originate "transferlist"
+      (Transferlist.Storage mempty mempty)
+      Transferlist.transferlistContract
```
pasqu4le, line 49: It looks like this used to be a simple address made with `createRole`:
```haskell
  (transferlistAddr, transferlist) <- createRole "transferlist"
```
which is a bit odd, any idea of why that was the case?

There is now also [a failure of this test](https://buildkite.com/serokell/stablecoin/builds/2268#0183eb5d-3c59-4403-947b-61adacbabf81), which doesn't seem related (it happens on the `deploy` following AFAICT) but perhaps it is :thinking:

haskell/src/Stablecoin/Client/Cleveland/IO.hs
```diff
@@ -108,22 +109,25 @@ naturalParser = decimal
 textParser :: Parser Text
 textParser = fromString <$> some printChar
 
+contractAddressParser :: Parser ContractAddress
+contractAddressParser = mkAddressParser parseKindedAddress
```
pasqu4le, line 113: Perhaps this function (and maybe `mkAddressParser` itself too) is worth upstreaming as well

haskell/test/Lorentz/Contracts/Nettest/StablecoinClientTest.hs
dcastro: > which is a bit odd, any idea of why that was the case?

I think it may have just been an oversight on my part a long time ago :sweat_smile: 

But yea, the introduction of kinded addresses helped highlight this mistake :tada: 

> There is now also [a failure of this test](https://buildkite.com/serokell/stablecoin/builds/2268#0183eb5d-3c59-4403-947b-61adacbabf81),

Seems to be related to the change made in this comment: https://github.com/tqtezos/stablecoin/pull/198#discussion_r998097985

Not sure why `refillable` didn't work as expected, I'll have a deeper look into this.

haskell/src/Stablecoin/Client/Cleveland/IO.hs
dcastro: Sure, but I think that seems a bit odd, `Morley.Michelson.Parser.Value` doesn't export any megaparsec parser for any typed michelson type (`Address`, `Mutez`, `MText`, etc) :thinking: We only have parsers for untyped values (e.g. `stringLiteral`), which we typecheck later.

Should we add megaparsec parsers for all of these?

dcastro: Personally, I don't think we should add something like this to morley, because the exact parser implementation depends on the context.

If you're parsing an address separated by spaces/periods (e.g. if the input string is `My address is tz1etc.`), then the parser should probably consume all alphanumeric characters, and then feed them into `Morley.Tezos.Address.parseAddress`.

But if you're parsing an address delimited by quotes (e.g. `PUSH address "tz1etc"`) then you probably want the parser to consumer all characters until it finds a `"`, and then feed them into `Morley.Tezos.Address.parseAddress`.

There is no generic one size fits all solution :/

ChangeLog.md
dcastro: I was going to do that in a separate follow-up PR ^^

Just pushed a few new commits to bump up the version number, update the costs, and update the instructions to generate the costs table (since the Kathmandu + lima faucets work differently now).

---

I also took the opportunity to add a "release checklist" to the PR template. Since every project has a slightly different release process, and it's hard to remember every single one, I've been adding these checklist to some of our int-dev projects, here are a few examples:
* [o-clock](https://github.com/serokell/o-clock/blob/master/.github/pull_request_template.md#-release-checklist)
* [xrefcheck](https://github.com/serokell/xrefcheck/blob/master/.github/pull_request_template.md#-release-checklist)

Let me know what you think.

dcastro: By the way, it seems the gas costs have gone up :eyes:

haskell/test/Lorentz/Contracts/Nettest/StablecoinClientTest.hs
dcastro: So, I just realized this doesn't work (see https://github.com/tqtezos/stablecoin/pull/198#discussion_r998515700), and it's because the `deploy` command below doesn't originate the contract via Cleveland.

It [actually calls the `stablecoin-client deploy`](https://github.com/tqtezos/stablecoin/blob/bc167c9d043dbcb0d95c297e0b73fde7cf314ee4/haskell/src/Stablecoin/Client/Cleveland/StablecoinImpl.hs#L99-L112) executable, which in turn will originate the contract via morley-client.

So the "refillable" mechanism is bypassed completely.

So our only options are to (1) transfer 5 tz up front, or (2) use cleveland's internals and use the moneybag account directly. Both have their pros and cons... I'm honestly ok with either one.

haskell/test/Lorentz/Contracts/Nettest/StablecoinClientTest.hs
dcastro: The failing test is fixed now

haskell/test/Lorentz/Contracts/Nettest/StablecoinClientTest.hs
lierdakil: Ah, I see. Ran into a similar problem with tzbtc smoke tests, with the same resolution. As for using the moneybag vs hard-coded amount, the former is more robust wrt changes in fees. I don't really have a decisive argument for one or the other, either.

dcastro: > the former is more robust wrt changes in fees

Yup, agreed, I ended up doing just that.

haskell/test/Lorentz/Contracts/Test/Management.hs
```diff
@@ -1023,14 +1020,14 @@ managementSpec originate =
...
-              originate $
+              originateSc $
                 ( defaultOriginationParams
                     (#owner :! testOwner)
                     (#pauser :! testPauser)
                     (#masterMinter :! testMasterMinter)
                 )
             do
-              mgmNotContractOwner $ withSender testMasterMinter $ call stablecoinContract (Call @"Change_master_minter") wallet1
+              mgmNotContractOwner $ withSender testMasterMinter $ transfer stablecoinContract $ calling (ep @"Change_master_minter") $ toAddress wallet1
```
lierdakil, line 1026: Could you wrap some of these overly long lines?

haskell/test/SMT.hs
```diff
@@ -176,45 +175,45 @@ resultToSs sval =
...
-  , ssOwner = rOwner $ sRoles storage
-  , ssMasterMinter = rMasterMinter $ sRoles storage
-  , ssPauser = rPauser $ sRoles storage
-  , ssPendingOwner = rPendingOwner $ sRoles storage
+  { ssMintingAllowances = Map.mapKeys addressToImplicitAddress $ sMintingAllowances storage
+  , ssLedger = Map.mapKeys addressToImplicitAddress $ bmMap $ sLedger storage
+  , ssOwner = addressToImplicitAddress $ rOwner $ sRoles storage
+  , ssMasterMinter = addressToImplicitAddress $ rMasterMinter $ sRoles storage
+  , ssPauser = addressToImplicitAddress $ rPauser $ sRoles storage
+  , ssPendingOwner = addressToImplicitAddress <$> rPendingOwner (sRoles storage)
```
lierdakil, line 191: I'm not sure if it makes sense, but this bunch of unsafe coercions seems kind of suspicious.

haskell/test/Lorentz/Contracts/Test/Common.hs
```diff
@@ -149,7 +149,12 @@ constructSingleTransfer
 constructSingleTransfer (arg #from_ -> from) (arg #to_ -> to) (arg #amount -> amount)
     = [TransferItem from [TransferDestination to FA2.theTokenId amount]]
 
-type OriginationFn param st m = OriginationParams -> m (ContractHandle param st ())
+originateStablecoin :: MonadCleveland caps m => OriginationParams -> m (ContractHandle SC.Parameter SC.Storage ())
+originateStablecoin originationParams = do
+  originate
+    "Stablecoin contract"
+    (mkInitialStorage originationParams)
+    stablecoinContract
```
lierdakil, line 157: Redundant `do`

haskell/src/Stablecoin/Client/Impl.hs
```diff
@@ -66,11 +66,14 @@ import Lorentz.Contracts.Stablecoin
...
-import Stablecoin.Client.Contract (InitialStorageData(..), mkInitialStorage)
+import Stablecoin.Client.Contract
+  (InitialStorageData(..), InitialStorageOptions(..), mkInitialStorage)
+import Stablecoin.Client.L1AddressOrAlias
+import Stablecoin.Client.Metadata (ViewParam(..), callOffChainView)
 import Stablecoin.Client.Parser (ContractMetadataOptions(..))
 
 -- | An address and an optional alias, if one is found.
-data AddressAndAlias = AddressAndAlias Address (Maybe Alias)
+data AddressAndAlias = AddressAndAlias Address (Maybe Text)
```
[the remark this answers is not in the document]
dcastro: Yup, I was going with that, but that would make implementing [this megaparsec parser](https://github.com/tqtezos/stablecoin/blob/bc167c9d043dbcb0d95c297e0b73fde7cf314ee4/haskell/src/Stablecoin/Client/Cleveland/IO.hs#L125-L129) harder.

I mean it's do-able, we'd have to (1) check the kind of the address that we just parsed, and then wrap the text in either `ImplicitAlias` or `ContractAlias`, and (2) fail the parser if we get a transaction rollup address. But harder.

So I dug a bit to see how this type is actually used:
1. In the `stablecoin-client` executable, we return it and then [print it to stdout immediately after](https://github.com/tqtezos/stablecoin/blob/bc167c9d043dbcb0d95c297e0b73fde7cf314ee4/haskell/src/Stablecoin/Client/Impl.hs#L290-L294).
2. In `Stablecoin.Client.Cleveland` (some sort of crossing between Cleveland and stablecoin-client), we [use it to parse an alias from stdout](https://github.com/tqtezos/stablecoin/blob/bc167c9d043dbcb0d95c297e0b73fde7cf314ee4/haskell/src/Stablecoin/Client/Cleveland/StablecoinImpl.hs#L83). However, the tests that call these functions [don't really look at the aliases though](https://github.com/tqtezos/stablecoin/blob/bc167c9d043dbcb0d95c297e0b73fde7cf314ee4/haskell/test/Lorentz/Contracts/Nettest/StablecoinClientTest.hs#L124-L125), they only check that the address is correct :thinking: (I think they should check the alias, but that's neither here nor there).

I concluded that keep tracking of the kind of the alias wasn't really worth it here, IMHO.

It's not an opinion I hold strongly though, so if you think we should use `Alias kind`, I can do it no problem ^^

haskell/src/Stablecoin/Client/Impl.hs
```diff
@@ -134,228 +142,213 @@ deploy (arg #sender -> sender) alias initialStorageData@InitialStorageData {..}
...
   transferlistAddr <- traverse resolveAddress transferlist
-  call sender contract (Call @"Set_transferlist") transferlistAddr
+  call sender contract (Call @"Set_transferlist") (toAddress <$> transferlistAddr)
 
 -- | Retrieve the contract's balance.
-getBalance :: "contract" :! AddressOrAlias -> MorleyClientM Mutez
+getBalance :: "contract" :! ContractAddressOrAlias -> MorleyClientM Mutez
 getBalance (arg #contract -> contract) = do
-  contractAddr <- resolveAddress contract
+  contractAddr :: ContractAddress <- resolveAddress contract
```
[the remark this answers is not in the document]
dcastro: Oh seems like a remnant from some temporary code changes, I'll delete it.

haskell/src/Stablecoin/Client/Impl.hs
```diff
@@ -378,20 +371,26 @@ getContractMetadata contract = do
 throwMdErr :: (a -> Text) -> Either a b -> MorleyClientM b
 throwMdErr toMsg f = either (throwM . SCEMetadataError . toMsg) pure f
 
+addressToL1Address :: Text -> Address -> MorleyClientM L1Address
+addressToL1Address fieldName = \case
```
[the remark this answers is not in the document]
dcastro: Dunno, I usually associate `unsafe*` functions with something that throws an exception in pure code with e.g. `error` :thinking: But I do see your point about making the "unsafeness" more obvious here, I'll add the prefix.

haskell/src/Stablecoin/Client/Impl.hs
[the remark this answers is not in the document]
lierdakil: > implementing this megaparsec parser harder.

Not that much harder though? From the top of my head,

```haskell
do
  addr <- labelled (label <> " address") addressParser
  mbAlias <- fmap mkAlias <$> optional (P.try (labelled (label <> " alias") aliasParser)) \\ addressKindSanity addr
  pure $  AddressAndAlias addr mbAlias
```

ChangeLog.md
pasqu4le: the release checklist sounds good to me, I think you included everything needed there :ok_hand: 

I mean, I guess it's still possible to miss some things to update in the README, but those are all in the same place and making a more detailed checklist seems unpractical.

> it seems the gas costs have gone up

indeed, transaction costs seem to be the same, but gas has gone up considerably.

There is some difference in the resulting Michelson contract (aside from formatting, I mean), but at a quick look I don't think that they account for it.

I should mention that there is some variation between calls and that there might have been changes in the gas model too. TBH I don't think that these changes in costs are concerning, hence why I didn't look at these too deeply,

haskell/src/Stablecoin/Client/Cleveland/IO.hs
pasqu4le: mmmmmh, yeah, you are right, those are fair points :ok_hand: 

TBH I think that for a moment I was thinking of options parsers for CLI with `optparse-applicative` and got confused here.

Those we do cover more extensively (which makes sense to me) and we already have parsers for different address types there, so no need to add more :ok_hand:

pasqu4le (commented):
I have no additional comments, I'll let @lierdakil check that his are addressed too and let him approve :slightly_smiling_face:

haskell/src/Stablecoin/Client/Impl.hs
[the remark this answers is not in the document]
dcastro: Done

haskell/src/Stablecoin/Client/Impl.hs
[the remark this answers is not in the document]
dcastro: Done

haskell/test/Lorentz/Contracts/Test/Management.hs
dcastro: Done. Also removed some redundant `do` blocks, and fused usages of `withSender`.

haskell/test/Lorentz/Contracts/Test/Common.hs
dcastro: Deleted

haskell/test/SMT.hs
dcastro: I think it's fine here - these tests keep converting between `Storage` and `SimpleStorage`, back and forth. The former contains `Address`es, the latter contains `ImplicitAddress`es.

BTW I added the `unsafe*` prefix here as well.

haskell/src/Stablecoin/Client/Contract.hs
```diff
@@ -2,43 +2,46 @@
...
+  { isoMasterMinter :: L1AddressOrAlias
+  , isoContractOwner :: L1AddressOrAlias
+  , isoPauser :: L1AddressOrAlias
+  , isoTransferlist :: Maybe ContractAddressOrAlias
+  , isoTokenSymbol :: Text
+  , isoTokenName :: Text
+  , isoTokenDecimals :: Natural
+  , isoContractMetadataStorage :: ContractMetadataOptions
+  , isoDefaultExpiry :: Expiry
+  }
```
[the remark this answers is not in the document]
dcastro: I have... some thoughts on DRY :sweat_smile: 

I see it as a means to an end; not as an end itself.

IMHO, DRY is not strictly about reducing the SLOC count.
The point of DRY is to avoid situations where you have the same (business, domain, whatever) logic duplicated across the codebase, the requirements change, and now you have to update multiple places. If you update some but forget others, you now have a bug.

In another sense, DRY is also useful when designing a public API: you don't want to have multiple ways of doing the same thing, that makes it way harder to use and remember. Homogeneity and consistency are great.

IMO, neither of these 2 concerns apply in this situation.

--- 

More importantly, these 2 data types are not _conceptually_ the same thing.
They're only _accidentally_ similar.

`InitialStorageOptions` is where we store the data we need to pass to the `deploy` command, and `InitialStorageData` is the where we store the data to create the `Storage`.

They _happen_ to have the same number of fields, with similar types, but they don't necessarily have to.
It's possible `deploy` could do some work to "crunch" 3 fields from `InitialStorageOptions` into a single field.

--- 

The problem with creating abstractions to encapsulate _accidental_ similarities is that, later, when we realize they should be 2 separate concepts, developers will be deterred from deleting the abstraction.

Going with the example from above: Say we create an abstraction `InitialStorageInfo stage`. If, later, we need to change `deploy` to crunch 3 fields into a single field, the developer will face 2 options:
1. Split `InitialStorageInfo stage` into two distinct data types, one with `n` fields and the other with `n+2` fields.
2. Move the crunching logic somewhere else, possibly to `mkInitialStorage`?

I think most devs would be inclined to go with (2) to avoid deleting the abstraction, even when (1) would be the best solution.


----


There are even more lax schools of thought like the [Rule of three](https://en.m.wikipedia.org/wiki/Rule_of_three_(computer_programming)), which I'm a bit on the fence about.

I do see the point of the rule, to avoid premature misguided abstractions, but I also think you can do that _without_ waiting for the pattern to emerge thrice.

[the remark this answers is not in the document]
dcastro: That being said, I just realized `InitialStorageData` and `mkInitialStorage` are only used in exactly one place, so I just inlined them :rofl:

haskell/src/Stablecoin/Client/L1AddressOrAlias.hs
```diff
@@ -0,0 +1,496 @@
...
+type ImplicitAddressOrAlias = KindedAddressOrAlias 'AddressKindImplicit
+
+----------------------------------------------------------------------------
+-- Reader
+----------------------------------------------------------------------------
+
+addressOrAliasKindSanity :: forall kind. KindedAddressOrAlias kind -> Dict (L1AddressKind kind, SingI kind)
+addressOrAliasKindSanity = \case
+  KAOAAddress addr -> Dict \\ addressKindSanity addr
+  KAOAAlias alias -> aliasKindSanity alias
```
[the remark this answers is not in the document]
dcastro: Oh I didn't know about `ArgDict` or `constraint-extras`! Really cool, thanks for the tip! :+1:

haskell/test/SMT.hs
dcastro: Ah I see your point! Yeah that would be a bigger effort ^^

Frankly, the stablecoin tests could use a deep refactoring. In particular, [these tests](https://github.com/tqtezos/stablecoin/blob/bc167c9d043dbcb0d95c297e0b73fde7cf314ee4/haskell/test/Lorentz/Contracts/Nettest/Nettest.hs) are a remnant from old times.

Before Cleveland was a thing, we used to write the vast majority of our tests via the emulator only, and then have a handful of expensive "nettest" network tests, just as a sort of smoke tests.

But now that _all_ the tests are written with Cleveland and are run on both the emulator and the network, these smoke tests are probably redundant.

Also, we didn't have any way of splitting network tests into individual tasty tests, so you can see [they're still all run as one](https://github.com/tqtezos/stablecoin/blob/bc167c9d043dbcb0d95c297e0b73fde7cf314ee4/haskell/test/Lorentz/Contracts/Nettest/Nettest.hs#L384-L389) :sweat_smile:

haskell/src/Stablecoin/Client/Contract.hs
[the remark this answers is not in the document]
lierdakil: Okay, all good points, and inlining is a better solution anyway :+1: 

That said, I could argue this similarity is not exactly accidental, but that's beside the point now, so I won't :laughing:

